PR #4, opened 2015-03-10
Start initial work on a Python 3 port

paultag:
This just gets the tests passing, no further work in validating the bindings on Python 3 was done yet.

Interesting notes - since `__getitem__` raises a `KeyError`, not an `AttributeError`, `hasattr` would bomb out. Neat.

This also eagerly consumes the `map` (since it's -- rightly -- an iterator), and does something similar with `.items` (which will be an iterator in Python 3 - I opted to not do two codepaths there, since the gain looks to be minimal usually)

### DIFF
--- a/inbox/client/client.py
+++ b/inbox/client/client.py
@@ -185,7 +185,7 @@ class APIClient(json.JSONEncoder):
 
         response = self._get_http_session(cls.api_root).get(url)
         results = _validate(response).json()
-        return map(lambda x: cls.create(self, namespace, **x), results)
+        return list(map(lambda x: cls.create(self, namespace, **x), results))
 
     @inbox_excepted
     def _get_resource_raw(self, namespace, cls, id, **filters):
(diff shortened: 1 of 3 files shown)


spang:
Cool! If you can sign our CLA, I'm happy to merge this: https://nilas.com/cla.html

We should consider adding a tox config file for Python 2.7 and 3.whatever, otherwise we're sure to break support at some point.

paultag:
I'll send in another PR tonight; CLA signed

On Tue, Mar 10, 2015 at 3:16 PM, Christine Spang <notifications@github.com>
wrote:

> Cool! If you can sign our CLA, I'm happy to merge this:
> https://nilas.com/cla.html
>
> We should consider adding a tox config file for Python 2.7 and 3.whatever,
> otherwise we're sure to break support at some point.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/inboxapp/inbox-python/pull/4#issuecomment-78126648>.
>



-- 
:wq

PR #6, opened 2015-06-22
Fixed setup.py version import bug

veered:
Before changing this, `pip install nylas` would throw the following error:

Collecting nylas
  Downloading nylas-0.2.7.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 20, in <module>
      File "/private/var/folders/0v/n_kwg1kd41z153jfhcgktbrw0000gn/T/pip-build-s1YBGZ/nylas/setup.py", line 6, in <module>
        from _version import __VERSION__
    ImportError: cannot import name __VERSION__
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/0v/n_kwg1kd41z153jfhcgktbrw0000gn/T/pip-build-s1YBGZ/nylas

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ import sys
 
 from setuptools import setup, find_packages
 sys.path.append('inbox/')
-from _version import __VERSION__
+from inbox._version import __VERSION__
 
 
 def main():


jennielees:
So, we changed from this module style import a while back (in 0186ad3027b86c64c0829d17b73e6a743d13480e) as it was causing issues, which the sys.path line should've fixed. Though apparently not. I don't feel comfortable merging this immediately for this reason, but thanks for the PR - we'll take a look.

khamidou:
Thanks for your contribution. Unfortunately, this pull request has been obsoleted by 8b9f66991eb9a768415f9dcda78a287a92149979, so I'll have to close it.

Karim

PR #12, opened 2015-10-10
Updated client.py to support url lib error in Py3

TheShapGuy:
Python3 has moved the urlparse into a submodule, fixing that.


brettgerry:
Fixed in 7382e24f7534b94ce4d07cede5826875675090e9.

PR #21, opened 2016-01-18
added search capability for threads and messages

waqasjaved160:


pfista:
Thanks for taking the time to make this pull request! We ended up adding this feature in a slightly different way, so I'll close this now.

PR #23, opened 2016-02-05
Fixes a bug in using offset parameter queries

berndverst:
Offset is always set to 0 by default. This handles the case where an explicit offset is provided and the **items()** method is called on a Restful Collection.

```
from nylas import APIClient

client = APIClient(None, None, "XXXXX")

for thread in client.threads.where({"in": "Nylas", "limit": 1, "offset": 0}).items():
  print(thread.id)
```

- Verified providing no offset param, and explicitly setting offset to 0 produce the same result.
- Verified that any other offset value result in the corresponding message with correct offset.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,8 +340,8 @@ We have a three-step process for releasing a new version of the Python SDK. Reme
    Run the tests:
    
    ```shell
-   PYTHONPATH=/your-sdk-path python tests/oauth.py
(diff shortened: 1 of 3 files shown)


khamidou:
LGTM, thanks for fixing!

bengotow:
:+1:

khamidou:
@bengotow, do you poll github continuously?

grinich:
uthqkqd

PR #26, opened 2016-03-23
Switch from HTTP Basic auth to Authorization: Bearer

zarvox:
Conveniently, we can drop some of the bytes/str special-casing that was needed
previously to interact with b64encode under python3.


berndverst:
Looks great. Thanks for adding!

PR #33, opened 2016-10-20
Empty dicts are falsy.  Non-empty dicts are truthy.

mgilson:
This makes statements like the following inefficient -- particularly on python2.x where `dict.keys()` needs to build a new list.

```py
if len(some_dict.keys()) > 0:
   ...
```


pfista:
Thanks for taking the time to submit this PR! Could you please sign this CLA so that I can merge the changes in? https://nylas.com/cla @mgilson

mgilson:
@pfista -- I believe that I signed that yesterday (and the presubmit check is passing).  Would you like me to fill it out again?

PR #38, opened 2016-10-26
Add message_id to Event NylasAPIObject

mgilson:
The `message_id` field seems to be missing from the documentation, but it seems that it is present when the event was created via email and it is not present with the event is a true calendar event.  With this parameter, users can filter out "Invite"/"Accepted" events that all reference the same actual calendar event.


mgilson:
FWIW, this commit addresses [my questions](https://nylas-chat.slack.com/archives/api-help/p1477328282000496) in the nylas community slack channel.

PR #49, opened 2017-07-06
Pylint

singingwolfboy:
Use [pylint](https://pylint.org/) for linting.

This pull request depends on #44, and it also includes the changes from #48.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -1,0 +1,445 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code
+extension-pkg-whitelist=
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=examples
+
+# Add files or directories matching the regex patterns to the blacklist. The
+# regex matches against base names, not paths.
+ignore-patterns=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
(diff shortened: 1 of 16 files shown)


pylintrc
```diff
@@ -0,0 +1,445 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code
+extension-pkg-whitelist=
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=examples
```
singingwolfboy: This pull request ignores the `examples` directory for now, since they're probably going to be refactored up later on. It doesn't make sense to invest the effort to clean up code that we're just going to radically change.

setup.py
```diff
@@ -1,23 +1,28 @@
...
+import re
 from setuptools import setup, find_packages
 from setuptools.command.test import test as TestCommand
-from _client_sdk_version import __VERSION__
+
+
+VERSION = ''
+with open('nylas/_client_sdk_version.py', 'r') as fd:
+    VERSION = re.search(r'^__VERSION__\s*=\s*[\'"]([^\'"]*)[\'"]',
+                        fd.read(), re.MULTILINE).group(1)
```
singingwolfboy: This code allows us to avoid modifying `sys.path`, which is generally an anti-pattern. You can verify that the setup.py file still gets the same version number by running `python setup.py --version`.

PR #53, opened 2017-07-06
remove unneeded generate_id() function

singingwolfboy:


nylas/client/client.py
```diff
@@ -145,8 +145,7 @@ def authentication_url(self, redirect_uri, login_hint=''):
                 'client_id': self.app_id,
                 'response_type': 'code',
                 'scope': 'email',
-                'login_hint': login_hint,
-                'state': generate_id()}
+                'login_hint': login_hint}
```
singingwolfboy: I'd really rather not. I want to focus on getting to the first checkpoint of this work, which is Python 3 support (done) and 95+% test coverage (in progress). We agreed that there would be no changes in functionality until the first checkpoint is done.

If this is an important feature to add, maybe we can make a note of it and come back later?

pfista: We shouldn't remove the state param yet, so we could just pass an empty string for now until we add support for the argument in a future PR.

singingwolfboy:
@pfista I just pushed another commit to this pull request that does that.

PR #62, opened 2017-07-17
Increase test coverage

singingwolfboy:
Also fixes some bugs around the `client.upgrade()` and `client.downgrade()` logic, as well as Python 3 `str` vs `bytes` complexities

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -80,7 +80,7 @@ def mock_account(api_url, account_id):
             "object": "account",
             "provider": "gmail",
             "organization_unit": "label",
-            "billing_state": "trial",
+            "billing_state": "paid",
         }
     )
     responses.add(
@@ -104,7 +104,7 @@ def mock_accounts(api_url, account_id, app_id):
             "object": "account",
             "provider": "gmail",
             "organization_unit": "label",
-            "billing_state": "trial",
+            "billing_state": "paid",
         }
     ])
     url = "{base}/a/{app_id}/accounts".format(base=api_url, app_id=app_id)
@@ -891,7 +891,7 @@ def mock_events(api_url):
 
 
 @pytest.fixture
-def mock_upgrade(api_url, account_id, app_id):
+def mock_account_management(api_url, account_id, app_id):
     account = {
         "account_id": account_id,
         "email_address": "ben.bitdiddle1861@gmail.com",
@@ -900,11 +900,11 @@ def mock_upgrade(api_url, account_id, app_id):
         "object": "account",
         "provider": "gmail",
         "organization_unit": "label",
-        "billing_state": "trial",
+        "billing_state": "paid",
     }
-    trial_response = json.dumps(account)
-    account["billing_state"] = "paid"
     paid_response = json.dumps(account)
+    account["billing_state"] = "cancelled"
+    cancelled_response = json.dumps(account)
 
     upgrade_url = "{base}/a/{app_id}/accounts/{id}/upgrade".format(
         base=api_url, id=account_id, app_id=app_id,
@@ -914,15 +914,15 @@ def mock_upgrade(api_url, account_id, app_id):
     )
     responses.add(
         responses.POST,
-        re.compile(upgrade_url),
+        upgrade_url,
         content_type='application/json',
         status=200,
         body=paid_response,
     )
     responses.add(
         responses.POST,
-        re.compile(downgrade_url),
+        downgrade_url,
         content_type='application/json',
         status=200,
-        body=trial_response,
+        body=cancelled_response,
     )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -43,6 +43,11 @@ def account_id():
 
 
 @pytest.fixture
+def app_id():
+    return 'fake-app-id'
+
+
+@pytest.fixture
 def api_client(api_url):
     return APIClient(None, None, None, api_url)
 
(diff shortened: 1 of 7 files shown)


tests/test_client.py
```diff
@@ -0,0 +1,120 @@
...
+    return (
+        url1.without_query() == url2.without_query() and
+        url1.query_dict == url2.query_dict
+    )
+
+
+def test_custom_client():
+    # Can specify API server
+    custom = APIClient(api_server="http://example.com")
+    assert custom.api_server == "http://example.com"
```
singingwolfboy: The exception message says that you must use `https://`, but doesn't actually require it...

tests/conftest.py
```diff
@@ -74,7 +79,8 @@ def mock_account(api_url, account_id):
             "name": "Ben Bitdiddle",
             "object": "account",
             "provider": "gmail",
-            "organization_unit": "label"
+            "organization_unit": "label",
+            "billing_state": "trial",
```
pfista: Can you change this so the default is `paid` please? This notion of trial will be deprecated soon so I'd like to avoid using it where possible (we're moving towards `paid`, `cancelled`, and `deleted` only). `paid` is the default state for all accounts, while downgrading will move them to `cancelled`. Sorry you didn't know this ahead of time :/

tests/conftest.py
```diff
@@ -88,6 +94,30 @@ def mock_account(api_url, account_id):
...
+    response_body = json.dumps([
+        {
+            "account_id": account_id,
+            "email_address": "ben.bitdiddle1861@gmail.com",
+            "id": account_id,
+            "name": "Ben Bitdiddle",
+            "object": "account",
+            "provider": "gmail",
+            "organization_unit": "label",
+            "billing_state": "trial",
```
pfista: `paid` please. See https://github.com/nylas/nylas-python/pull/62/files#diff-484462fced51d1a06b1d93b4a44dd535R83

tests/conftest.py
```diff
@@ -851,3 +888,41 @@ def mock_events(api_url):
...
+def mock_upgrade(api_url, account_id, app_id):
+    account = {
+        "account_id": account_id,
+        "email_address": "ben.bitdiddle1861@gmail.com",
+        "id": account_id,
+        "name": "Ben Bitdiddle",
+        "object": "account",
+        "provider": "gmail",
+        "organization_unit": "label",
+        "billing_state": "trial",
```
pfista: Can you change this to `cancelled` or `deleted` please? See https://github.com/nylas/nylas-python/pull/62/files#diff-484462fced51d1a06b1d93b4a44dd535R83

tests/conftest.py
```diff
@@ -851,3 +888,41 @@ def mock_events(api_url):
         status=200,
         body=response_body
     )
+
+
+@pytest.fixture
+def mock_upgrade(api_url, account_id, app_id):
```
pfista: Another nitpick: could you rename this to something like `mock_account_management` since it covers both upgrading and downgrading?

tests/conftest.py
```diff
@@ -851,3 +888,41 @@ def mock_events(api_url):
...
+        "account_id": account_id,
+        "email_address": "ben.bitdiddle1861@gmail.com",
+        "id": account_id,
+        "name": "Ben Bitdiddle",
+        "object": "account",
+        "provider": "gmail",
+        "organization_unit": "label",
+        "billing_state": "trial",
+    }
+    trial_response = json.dumps(account)
```
pfista: Similarly this should be `cancelled_response`.

tests/conftest.py
```diff
@@ -851,3 +888,41 @@ def mock_events(api_url):
...
+        content_type='application/json',
+        status=200,
+        body=paid_response,
+    )
+    responses.add(
+        responses.POST,
+        re.compile(downgrade_url),
+        content_type='application/json',
+        status=200,
+        body=trial_response,
```
pfista: `cancelled_response`

tests/test_accounts.py
```diff
@@ -21,15 +22,16 @@ def test_account_json(api_client, monkeypatch):
...
-    account.upgrade()
-    assert account.billing_state is True
-    account.downgrade()
-    assert account.billing_state is False
+@responses.activate
+@pytest.mark.usefixtures("mock_accounts", "mock_upgrade")
+def test_account_upgrade(api_client, app_id):
+    api_client.app_id = app_id
+    account = api_client.accounts.first()
+    assert account.billing_state == "trial"
```
pfista: `cancelled` here please!

tests/test_accounts.py
```diff
@@ -21,15 +22,16 @@ def test_account_json(api_client, monkeypatch):
...
+    account = account.upgrade()
+    assert account.billing_state == "paid"
+    account = account.downgrade()
+    assert account.billing_state == "trial"
```
pfista: `cancelled` here please!

tests/test_client.py
pfista: Could you create a ticket for that please?

singingwolfboy: Created #65

tests/conftest.py
singingwolfboy: What happens if you call `.upgrade()` on an account in `paid` status? What happens if you call `.downgrade()` on an account in `cancelled` or `deleted` status?

PR #67, opened 2017-07-17
Switch to using urlobject library

singingwolfboy:
Using a dedicated library is nicer than copy-pasting bits of Tornado. [URLObject docs are here.](https://urlobject.readthedocs.io/en/latest/index.html)


tests/test_client.py
```diff
@@ -60,7 +60,7 @@ def test_client_authentication_url(api_client, api_url):
             ('state', ''),
             ('redirect_uri', '/redirect'),
             ('response_type', 'code'),
-            ('client_id', 'None'),
+            ('client_id', ''),
```
singingwolfboy: I don't know if you consider this a bug or a feature, but it's a change to call out. The `url_concat()` function would transform `None` values into the string "None". URLObject is more intelligent than that.

nylas/client/util.py
```diff
@@ -1,30 +0,0 @@
-from six.moves.urllib.parse import urlencode
```
mhahnenberg: Hooray! 🎉

tests/test_client.py
mhahnenberg: Hmm, this is technically a breaking change. I know it's sorta weird, but would it be possible to go back to the old behavior and change the earlier line in `authentication_url` to `'client_id': self.app_id or 'None',`?

nylas/client/util.py
mhahnenberg: Hooray! 🎉

PR #72, opened 2017-07-27
Fix failing test

singingwolfboy:


tests/test_drafts.py
```diff
@@ -18,8 +17,8 @@ def test_save_send_draft(api_client):
     draft.save()
 
     draft.subject = "Stay polish, stay hungary"
-    draft.save(random_query='true', param2='param')
-    assert draft.subject == 'Update #2'
+    draft.save()
+    assert draft.subject == "Stay polish, stay hungary"
```
singingwolfboy: This was a case of "how did that _ever_ work?"

The test sets the subject to "Stay polish, stay hungary", then saves the draft, then asserts that the subject is _no longer_ that string, but is now "Update #2", for some reason. I changed the test to make more sense. I have no idea how it was ever passing to begin with.

PR #74, opened 2017-07-27
Add native auth example for Exchange

singingwolfboy:
login form

### DIFF
--- a/examples-new/native-authentication-exchange/config.json
+++ b/examples-new/native-authentication-exchange/config.json
@@ -1,0 +1,5 @@
+{
+  "SECRET_KEY": "replace me with a random string",
+  "NYLAS_OAUTH_API_ID": "replace me with the API ID from Nylas",
+  "NYLAS_OAUTH_API_SECRET": "replace me with the API Secret from Nylas"
+}
--- a/examples-new/native-authentication-exchange/config.json
+++ b/examples-new/native-authentication-exchange/config.json
@@ -1,5 +1,5 @@
 {
   "SECRET_KEY": "replace me with a random string",
-  "NYLAS_OAUTH_API_ID": "replace me with the API ID from Nylas",
-  "NYLAS_OAUTH_API_SECRET": "replace me with the API Secret from Nylas"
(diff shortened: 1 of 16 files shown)


examples-new/native-authentication-exchange/server.py
```diff
@@ -0,0 +1,166 @@
...
+    # Nylas needs in order to connect, and sending it to the authorize API.
+    nylas_authorize_data = {
+        "client_id": app.config["NYLAS_OAUTH_API_ID"],
+        "name": name,
+        "email_address": email,
+        "provider": "exchange",
+        "settings": {
+            "username": email,
+            "password": password,
+            "eas_server_host": server_host,
```
pfista: You should only send this value if it is not None

examples-new/native-authentication-exchange/config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "SECRET_KEY": "replace me with a random string",
+  "NYLAS_OAUTH_API_ID": "replace me with the API ID from Nylas",
```
pfista: Can you rename these to `NYLAS_CLIENT_ID` and `NYLAS_CLIENT_SECRET` please?

singingwolfboy: I named them that way for consistency with the other two examples. The "hosted-oauth" example uses those names to set up OAuth via Flask-Dance: https://flask-dance.readthedocs.io/en/latest/providers.html#module-flask_dance.contrib.nylas

Do you want me to make this example different from the others? Do you want me to use a different name in all of them? That will require pushing out a new version of Flask-Dance with that update, as well.

pfista: Could you use a different name in all of them please? On our end, we talk to customers in terms of a Nylas `client_id` and `client_secret`, so I'd like to keep that naming convention consistent. Sorry for not catching it earlier!

singingwolfboy: In that case, you should probably change it in your developer console, as well. That's where I got the "API ID" and "API Secret" terminology from. I assume you'd also like me to update it in the README files, and everywhere else I've used those terms?
nylas developer console

PR #77, opened 2017-08-02
Remove old examples

singingwolfboy:
Is there any reason not to get rid of the old examples?


singingwolfboy:
I'll reopen a similar pull request after #83 merges.

PR #80, opened 2017-08-03
Ensure draft versions are updated

### LINKED ISSUE #61: Ensure draft versioning works
Make sure that draft versions are updated correctly and automatically when creating and updating drafts.

singingwolfboy:
Fixes #61

### DIFF
--- a/tests/test_drafts.py
+++ b/tests/test_drafts.py
@@ -62,3 +62,15 @@ def test_delete_draft(api_client):
     draft.save()
     # ... and delete it for real.
     draft.delete()
+
+
+@pytest.mark.usefixtures("mock_draft_saved_response")
+def test_draft_version(api_client):
+    draft = api_client.drafts.create()
+    assert 'version' not in draft
+    draft.save()
+    assert draft['version'] == 1
+    draft.save()
+    assert draft['version'] == 2
+    draft.save()
+    assert draft['version'] == 3
--- a/tests/test_drafts.py
+++ b/tests/test_drafts.py
@@ -69,8 +69,8 @@ def test_draft_version(api_client):
     draft = api_client.drafts.create()
     assert 'version' not in draft
     draft.save()
+    assert draft['version'] == 0
+    draft.update()
     assert draft['version'] == 1
-    draft.save()
+    draft.update()
     assert draft['version'] == 2
-    draft.save()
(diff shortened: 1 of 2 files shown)


tests/test_drafts.py
```diff
@@ -65,3 +65,16 @@ def test_delete_draft(api_client):
     draft.save()
     # ... and delete it for real.
     draft.delete()
+
+
+@responses.activate
+@pytest.mark.usefixtures("mock_draft_saved_response")
+def test_draft_version(api_client):
+    draft = api_client.drafts.create()
+    assert 'version' not in draft
```
pfista: Hmm, so if a draft is created with the API it's first version should start at 0. Is this not the case in the SDK?

singingwolfboy: The [`create()`](https://github.com/singingwolfboy/nylas-python/blob/master/nylas/client/restful_models.py#L26-L50) method will populate the object based on the `kwargs` passed into this function. So to create an object with an initial version of 0, you can do:

```python
draft = api_client.drafts.create(version=0)
```

However, there's no special handling for draft numbers in the `create()` method itself, so if you leave it unspecified, then it won't be in resulting Draft instance. Do you want me to change that, so that an unsaved Draft instance automatically has a version number of 0?

The draft numbers in this particular test are coming from the `mock_draft_saved_response` fixture. Right now, the fixture will increase the draft number by one every time. Do you want me to change the fixture so that the first save returns a version number of 0, instead of 1?

Note that if you want both of these modifications, that means that the first time you save a Draft instance, the version number _will not change_. Is that correct?

pfista: @singingwolfboy With the API, the first time a draft is `POST`ed it will return a version of 0 (and no version needs to be passed in to this `POST`). Later, if a PUT request is made on that draft, you must pass `"version": 0`. The json response of the updated draft will include `"version": 1`.

Since it looks like the `create()` method only creates a local copy of the object and doesn't actually make an API request until `save()` is called it, then after the first `save()` the draft version returned should be 0 according to the logic above.

Could you change the other draft saves to be `update()` instead to be a little more explicit please, and make sure the draft version is 0 after the first save?

singingwolfboy: Fixed.

PR #81, opened 2017-08-03
Refactor how client validates responses

### LINKED ISSUE #63: Handle HTTP success status codes beyond 200
There are [many HTTP status codes that indicate different kinds of successful responses](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_Success). "200 OK" is the most widely used, but many RESTful APIs use "201 Created" in response to creating resources, and "204 No Content" in response to deleting resources.

This project is hardcoded to only accept the 200 status code as success, which means that if the API ever returns a different, non-200 status code to indicate a successful response, the SDK will throw an exception. This project should work with success status codes beyond 200.

When using the `requests` library, you can use [`response.ok`](http://docs.python-requests.org/en/master/api/#requests.Response.ok) to determine if the response was successful. You can also use [`response.raise_for_status()`](http://docs.python-requests.org/en/master/api/#requests.Response.raise_for_status) to raise an exception if necessary, rather than trying to determine if the request was successful and raising an exception if it wasn't.

### LINKED ISSUE #64: Check for compatibility with redirect responses from the API
Sometimes, resources move: things are renamed or reorganized, or deleted and re-created. An API should be able to issue [redirect responses](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_Redirection) to clients seeking information about resources that have moved.

I don't know if this SDK works properly with redirected responses. By default, the `requests` library [will handle redirects for you](http://docs.python-requests.org/en/master/user/quickstart/#redirection-and-history), but we should verify that it does the right thing.

singingwolfboy:
This will also accept successful responses with non-200 status codes. For example, "201 Created" is a successful status code. Fixes #63 and fixes #64.

### DIFF
--- a/nylas/client/client.py
+++ b/nylas/client/client.py
@@ -14,6 +14,10 @@ from nylas.client.restful_models import (
     Label, Draft
 )
 from nylas.client.errors import APIClientError, ConnectionError, STATUS_MAP
+try:
+    from json import JSONDecodeError
+except ImportError:
+    JSONDecodeError = ValueError
 
 DEBUG = environ.get('NYLAS_CLIENT_DEBUG')
 API_SERVER = "https://api.nylas.com"
@@ -40,7 +44,7 @@ def _validate(response):
     try:
         data = response.json()
         json_content = True
-    except json.JSONDecodeError:
+    except JSONDecodeError:
         data = response.content
         json_content = False
 
--- a/nylas/client/client.py
+++ b/nylas/client/client.py
@@ -13,64 +13,57 @@ from nylas.client.restful_models import (
     Account, APIAccount, SingletonAccount, Folder,
     Label, Draft
 )
-from nylas.client.errors import (
-    APIClientError, ConnectionError, NotAuthorizedError,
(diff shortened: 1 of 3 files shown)


nylas/client/client.py
```diff
@@ -13,64 +13,61 @@
...
+    if response.status_code in STATUS_MAP:
+        cls = STATUS_MAP[response.status_code]
+        if json_content:
+            if "message" in data:
+                kwargs["message"] = data["message"]
+            if "server_error" in data:
+                kwargs["server_error"] = data["server_error"]
+            raise cls(**kwargs)
+        else:
+            kwargs["message"] = "Malformed"
```
pfista: Can you be a little more specific with this error message please? "Error decoding JSON response"

singingwolfboy: This was the existing error message. Can you confirm that you want me to change it from how it used to be?

pfista: Ah in that case we should avoid a potentially breaking change! Thanks

PR #83, opened 2017-08-03
Added webhook example

singingwolfboy:


examples-new/webhooks/server.py
```diff
@@ -0,0 +1,157 @@
...
+    if not is_genuine:
+        return "Signature verification failed!", 401
+
+    # Alright, we have a genuine webhook notification from Nylas!
+    # Let's find out what it says...
+    data = request.get_json()
+    for delta in data["deltas"]:
+        # This is the part of the code where you would process the information
+        # from the webhook notification. Each delta is one change that happened,
+        # and might require fetching message IDs, updating your database,
```
pfista: Can you mention that any processing or fetching should be done **async** please? We recommend that customers queue work and return a 200 to prevent Nylas POST from timing out. I believe our timeout is 5 minutes right now.

examples-new/webhooks/server.py
```diff
@@ -0,0 +1,157 @@
...
+    to figure out what URL it has assigned, and suggest that to the user.
+    https://ngrok.com/docs#list-tunnels
+    """
+    try:
+        ngrok_resp = requests.get("http://localhost:4040/api/tunnels")
+    except requests.ConnectionError:
+        # I guess ngrok isn't running.
+        return None
+    ngrok_data = ngrok_resp.json()
+    secure_urls = [
```
pfista: nice

examples-new/webhooks/templates/index.html
```diff
@@ -0,0 +1,12 @@
+{% extends "base.html" %}
+{% block body %}
+<p>This example doesn't have anything to see in the browser. Set up your
```
pfista: Can you link to Ngroks web interface please? It's nice to replay POSTs from Nylas this way without needing to actually re-create a triggered notification.

singingwolfboy:
@pfista I've integrated [Celery](http://www.celeryproject.org/) into the example project, so that the webhook notifications are processed in asynchronous tasks. This requires having a running [redis](https://redis.io/) server. If that's too difficult for the person trying the example, they can disable Celery by setting the [`CELERY_TASK_ALWAYS_EAGER`](http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-task_always_eager) config value to True.

examples-new/webhooks/templates/index.html
singingwolfboy: Done.

PR #88, opened 2017-08-07
Convert timestamp fields to datetime fields

### LINKED ISSUE #87: Filtering on booleans and datetimes fails
When trying to filter on boolean fields, like `starred` or `unread`, the SDK converts the boolean object to a string, so `True` becomes "True" and `False` becomes "False". Both of these strings are treated as true by the backend. As a result, it's not possible to filter on boolean False for any boolean fields.

```python
>>> client.threads.where(unread=True).all()
# fetches `https://api.nylas.com/threads?unread=True&offset=0&limit=50`
>>> client.threads.where(unread=False).all()
# fetches `https://api.nylas.com/threads?unread=False&offset=0&limit=50`
# same results either way
```

In addition, in Python it's common to use the [`datetime`](https://docs.python.org/3/library/datetime.html) module to refer to dates and times. There are several APIs that return datetimes, or accept datetimes as parameters (like `starts_before` when filtering events). The SDK _only_ accepts timestamps, and does not work with datetime objects at all. In fact, trying to pass a datetime object to the SDK results in an internal exception, like this:

```python
>>> now
datetime.datetime(2017, 8, 7, 16, 56, 36, 56763)
>>> client.events.where(starts_before=now).all()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/singingwolfboy/clones/nylas-python/nylas/client/restful_model_collection.py", line 45, in all
    return self._range(self.filters['offset'], limit)
  File "/Users/singingwolfboy/clones/nylas-python/nylas/client/restful_model_collection.py", line 118, in _range
    to_fetch)
  File "/Users/singingwolfboy/clones/nylas-python/nylas/client/restful_model_collection.py", line 107, in _get_model_collection
    **filters)
  File "/Users/singingwolfboy/clones/nylas-python/nylas/client/client.py", line 76, in caught
    return func(*args, **kwargs)
  File "/Users/singingwolfboy/clones/nylas-python/nylas/client/client.py", line 259, in _get_resources
    url = str(URLObject(url).add_query_params(filters.items()))
  File "/Users/singingwolfboy/.virtualenvs/nylas-python/lib/python3.6/site-packages/urlobject/urlobject.py", line 464, in add_query_params
    return self.with_query(self.query.add_params(*args, **kwargs))
  File "/Users/singingwolfboy/.virtualenvs/nylas-python/lib/python3.6/site-packages/urlobject/query_string.py", line 67, in add_params
    new = new.add_param(name, value)
  File "/Users/singingwolfboy/.virtualenvs/nylas-python/lib/python3.6/site-packages/urlobject/query_string.py", line 58, in add_param
    parameter = qs_encode(name) + '=' + qs_encode(value)
  File "/Users/singingwolfboy/.virtualenvs/nylas-python/lib/python3.6/site-packages/urlobject/query_string.py", line 138, in _qs_encode_py3
    return urlparse.quote_plus(s)
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/urllib/parse.py", line 791, in quote_plus
    string = quote(string, safe + space, encoding, errors)
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/urllib/parse.py", line 775, in quote
    return quote_from_bytes(string, safe)
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/urllib/parse.py", line 800, in quote_from_bytes
    raise TypeError("quote_from_bytes() expected bytes")
TypeError: quote_from_bytes() expected bytes
```

singingwolfboy:
Fixes #87.

This pull request will allow the SDK to automatically convert timestamp integers into datetime objects in Python. I wanted to make these datetime fields simply views of the timestamp data, but I figured it would be better to not introduce too much complexity into these models, especially since they're just `dict` subclasses to begin with. Check out the `test_update_thread_attrs` test for the implications of this.

Thoughts?

### DIFF
--- a/nylas/client/client.py
+++ b/nylas/client/client.py
@@ -258,7 +258,9 @@ class APIClient(json.JSONEncoder):
                 postfix
             )
 
-        converted_filters = convert_datetimes_to_timestamps(filters, cls.datetime_attrs)
+        converted_filters = convert_datetimes_to_timestamps(
+            filters, cls.datetime_filter_attrs,
+        )
         url = str(URLObject(url).add_query_params(converted_filters.items()))
         response = self._get_http_session(cls.api_root).get(url)
         results = _validate(response).json()
@@ -283,7 +285,9 @@ class APIClient(json.JSONEncoder):
             url = "{}/a/{}/{}/{}{}".format(self.api_server, self.app_id,
                                            cls.collection_name, id, postfix)
 
-        converted_filters = convert_datetimes_to_timestamps(filters, cls.datetime_attrs)
+        converted_filters = convert_datetimes_to_timestamps(
+            filters, cls.datetime_filter_attrs,
+        )
         url = str(URLObject(url).add_query_params(converted_filters.items()))
 
         response = self._get_http_session(cls.api_root).get(url, headers=headers)
--- a/nylas/client/client.py
+++ b/nylas/client/client.py
@@ -19,7 +19,7 @@ from nylas.client.restful_models import (
     Label, Draft
 )
 from nylas.client.errors import APIClientError, ConnectionError, STATUS_MAP
-from nylas.utils import timestamped_data
+from nylas.utils import convert_datetimes_to_timestamps
 
 DEBUG = environ.get('NYLAS_CLIENT_DEBUG')
 API_SERVER = "https://api.nylas.com"
@@ -258,8 +258,8 @@ class APIClient(json.JSONEncoder):
                 postfix
             )
 
-        ts_filters = timestamped_data(filters, cls.datetime_attrs)
-        url = str(URLObject(url).add_query_params(ts_filters.items()))
+        converted_filters = convert_datetimes_to_timestamps(filters, cls.datetime_attrs)
+        url = str(URLObject(url).add_query_params(converted_filters.items()))
         response = self._get_http_session(cls.api_root).get(url)
         results = _validate(response).json()
         return [
@@ -283,8 +283,8 @@ class APIClient(json.JSONEncoder):
             url = "{}/a/{}/{}/{}{}".format(self.api_server, self.app_id,
                                            cls.collection_name, id, postfix)
 
-        ts_filters = timestamped_data(filters, cls.datetime_attrs)
-        url = str(URLObject(url).add_query_params(ts_filters.items()))
+        converted_filters = convert_datetimes_to_timestamps(filters, cls.datetime_attrs)
+        url = str(URLObject(url).add_query_params(converted_filters.items()))
 
         response = self._get_http_session(cls.api_root).get(url, headers=headers)
         return _validate(response)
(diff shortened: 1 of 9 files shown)


nylas/utils.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    Convert a datetime to a timestamp.
+    https://stackoverflow.com/a/8778548/141395
+    """
+    delta = dt - epoch
+    # return delta.total_seconds()
+    return delta.seconds + delta.days * 86400
+
+
+def timestamped_data(data, datetime_attrs):
```
mhahnenberg: Could we rename this? It makes me think we're timestamping all data. Maybe like "convert_datetimes_to_timestamps" or something similar?

nylas/client/client.py
```diff
@@ -256,7 +258,8 @@ def _get_resources(self, cls, extra=None, **filters):
                 postfix
             )
 
-        url = str(URLObject(url).add_query_params(filters.items()))
+        ts_filters = timestamped_data(filters, cls.datetime_attrs)
```
mhahnenberg: ts_filters is a slightly confusing name here. It makes it seem like it's only timestamp filters. Maybe converted_filters or request_filters?

nylas/client/restful_models.py
```diff
@@ -83,6 +93,12 @@ class Message(NylasAPIObject):
              "account_id", "object", "snippet", "starred", "subject",
              "thread_id", "to", "unread", "starred", "_folder", "_labels",
              "headers"]
+    datetime_attrs = {
+        "received_at": "date",
+        # The following are used for filtering:
```
mhahnenberg: This is a little bit confusing. Would it make sense to have a separate class field just for converting filter-only attributes? E.g. something like datetime_filter_attrs?

singingwolfboy: We could, but they're treated exactly the same as the other datetime attributes. Do you think it's worth the duplication? I figured that a comment would make it less confusing, but if you think we need to go further and have a separate field, I can do that.

nylas/utils.py
singingwolfboy: Ah, good point. Fixed.

nylas/client/client.py
singingwolfboy: Fixed.

nylas/client/restful_models.py
mhahnenberg: I was thinking instead of using `cls.datetime_attrs` as the argument to `convert_datetimes_to_timestamps` above we could use `cls.datetime_filter_attrs` instead since that's the only place we care about these filter attrs. And when defining the class field we could extend `cls.datetime_attrs` with the additional filter attrs.

singingwolfboy: Fair point. Done.

PR #89, opened 2017-08-07
Add client.delete_account()

singingwolfboy:
Can't make it a method on the account object, because the account doesn't have a reference to the client object :(


singingwolfboy:
This is more complex than we thought. Closing for now.

PR #91, opened 2017-08-08
pin responses library

singingwolfboy:
because the latest version has a bug :(


singingwolfboy:
Bug fix is here: https://github.com/getsentry/responses/pull/168

PR #93, opened 2017-11-15
RestfulModelCollection.items method now checks against limit filter

FabioFleitas:
Without this PR, if you manually set a limit that is higher than `CHUNK_SIZE` you might do unnecessary API calls.

For example, if you have 100 threads, and your limit is 100, this will check that `100 < CHUNK_SIZE` and would continue to do another API call for the next page offset.

We could probably clean up other places in this file that `CHUNK_SIZE` is used since we now make `limit` default to `CHUNK_SIZE`, but I kept the changes minimal.

### DIFF
--- a/nylas/client/restful_model_collection.py
+++ b/nylas/client/restful_model_collection.py
@@ -61,6 +61,7 @@ class RestfulModelCollection(object):
         if filter:
             filters.update(filter)
         filters.setdefault('offset', 0)
+        filters.setdefault('limit', CHUNK_SIZE)
         collection = copy(self)
         collection.filters = filters
         return collection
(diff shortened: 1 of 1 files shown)


hallamoore:
Hi @FabioFleitas,

Thanks for taking the time to submit this! However, making multiple API calls is the intended behavior. This prevents us from loading too much data into memory at one time. It's not that big a deal if your limit is just 100, but if someone requests 5,000 objects, we definitely want to break that up. 

I'm gonna close this pull request, but let me know if you have other concerns about this

FabioFleitas:
Without this change though you can't change the `limit` which results in more API calls. I found that your service handles 100 & 1000 in basically the same amount of time, but having to paginate through 10x more requests does significantly slow down the application. Although the memory issue is fair, I think the user should decide what's the best way to handle that given their server specs.

We've been using this PR and #93 in production with great results that allows us to fetch all the data we need with the least amounts of requests against your system.

hallamoore:
I'm hesitant to automatically couple the chunk size with the limit, but what do you think of adding an additional `chunk_size` filter? This would allow more flexibility for users who would like to have a really large limit while still iterating over smaller chunks.

PR #94, opened 2017-11-16
Threads now have a messages attribute when passing view=expanded

FabioFleitas:


hallamoore:
Hi @FabioFleitas,

The Thread model actually already has a `messages` property that returns a Message collection (rather than just json), so you can get the individual Message models by doing something like this:

```
client = APIClient(app_id, app_secret, token)
thread = client.threads.find('ac123acd123ef123')
for message in thread.messages.items():
    # do something with message
```

I don't think adding support for the `view=expanded` parameter would improve this usage, so I'm going to close this pull request. Please feel free to further explain your use case if this is not sufficient

FabioFleitas:
Those are extra API calls, though. And `view=expanded` is built into your API so that you wouldn't have to do _N_ extra calls to fetch the messages.

Right now if someone does `view=expanded` with this Python library, it fetches the `messages`, but just discards them since it's not listed in the `attrs`. All this does is have it actually save the data that you are returning from your API.

hallamoore:
It's just one extra API call (all messages filtered by thread_id), but yeah that adds up. I think a couple changes need to be made before this will work as intended though. 

The attribute needs to be listed as `'_messages'` since `messages` is already defined on the Thread class (we already have logic in place to account for the underscore when populating the data from the API. `_folders` is another example of this). Then, the `messages` logic should be updated to check if the Thread has `_messages` defined. If it does, it can return those, if not, it should keep to the original behavior. 

(It'd also be awesome if you could add a test for the new code path)

hallamoore:
Oh, and like `folders`, `messages` should transform `_messages` into Message instances

singingwolfboy:
Unfortunately, I think we need to close this pull request at this point. It sounds like @FabioFleitas isn't interested in doing the work that @hallamoore asked for, so if we do it in the future, it can be in a different pull request.

PR #98, opened 2018-01-04
Upload files with declared filename, not OS filename

### LINKED ISSUE #95: File.filename doesn't apply when type is stream:
Setting a filename on an attachment only works when using `.data`. When `.stream` is used, it falls back to the name of the file on the uploading system.

```
        myfile = self.nylas_client.files.create()
        myfile.content_type = 'application/pdf'
        myfile.filename = attachment_name
        with open(attachment_path, 'rb') as f:
            myfile.stream = f
            myfile.save()

        myfile.filename = attachment_name
        # Create a new draft
        draft = self.nylas_client.drafts.create()
        if type(recipients) == str:
            recipients = [recipients]
        draft.to = [{'email': recipient} for recipient in recipients]
        draft.subject = subject
        draft.body = message
        draft.attach(myfile)
```

singingwolfboy:
Fixes #95.

### DIFF
--- a/nylas/client/restful_models.py
+++ b/nylas/client/restful_models.py
@@ -387,18 +387,27 @@ class File(NylasAPIObject):
     collection_name = 'files'
 
     def save(self):  # pylint: disable=arguments-differ
-        if hasattr(self, 'stream') and self.stream is not None:
-            data = {self.filename: self.stream}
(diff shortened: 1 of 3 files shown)


rcoh:
LGTM!

PR #100, opened 2018-01-09
Special fields for Contacts

singingwolfboy:
This pull request updates the Contact model to provide special handling for a few fields. Specifically, it parses the `birthday` field into a Python [date object](https://docs.python.org/3/library/datetime.html#date-objects), and parses the `email_addresses`, `im_addresses`, `physical_addresses`, `phone_numbers`, and `web_pages` fields into a [`MultiDict`](http://werkzeug.pocoo.org/docs/0.14/datastructures/#werkzeug.datastructures.MultiDict) object, imported from the Werkzeug library. If Werkzeug is not installed, it falls back on the standard Python dict, which cannot handle multiple values for a single key.

The automated tests show how these special fields can be used.


nylas/client/restful_models.py
```diff
@@ -1,17 +1,38 @@
...
 from six import StringIO
 
+try:
+    from werkzeug.datastructures import MultiDict
+except ImportError:  # pragma: no cover
+    MultiDict = dict
+
 # pylint: disable=attribute-defined-outside-init
 
+def detect_attr(value):
```
hallamoore: This function was bothering me a lot because it's named kinda vaguely named even though it does something very specific and we already have `datetime_attrs` lying around. I think the issue here is that we've made `birthday` a weird object that copied aspects from `Event.when`, even though `Event.when` is usually a time-range and not a single date. I've suggested that we change `birthday` to just a normal timestamp, similar to `Thread.first_message_timestamp`. If that goes through, then we can just add `birthday` to the `datetime_attrs` instead.

nylas/client/restful_models.py
```diff
@@ -1,17 +1,38 @@
...
+    attr_type = value.get("object", None)
+    if attr_type == "date":
+        date_str = value["date"]
+        if date_str:
+            return datetime.strptime(date_str, "%Y-%m-%d").date()
+    return None
+
+
+def mdict_attr(items, attr_name=None):
```
hallamoore: My takeaways from our discussion yesterday:
- If we use MultiDict, we should make the dependency mandatory. The fallback to dict would only keep the last value of each type, but 95% of our users should care about the full list for each type. 
- The only benefit of MultiDict over parsing our own dict is that MultiDict allows users to get a single value for a type.

I agree that the typed dict is useful, but I don't think the 'single value' functionality is useful enough to warrant introducing the MultiDict dependency/different syntax. I think we should parse the data into a normal dict, and rename `mdict_attr` to something like `typed_dict_attr`.

singingwolfboy:
@hallamoore I believe I've addressed your comments. Can you re-review?

PR #101, opened 2018-01-09
Delegate error handling to Requests library

singingwolfboy:


singingwolfboy:
@hallamoore I believe I've addressed your comment. Can you re-review?

PR #103, opened 2018-08-28
Raise UnSyncedError when message.raw GET request returns 202

jetzhou:
Within a few seconds of a message sent, `message.raw` will return with a 202 and content `Message {message_id} has not been synced yet`, so the caller will proceed as though that error message is the actual raw content. This change will raise an Error when that's the case so it's much easier for caller to be aware and deal with that situation.

### DIFF
--- a/nylas/client/errors.py
+++ b/nylas/client/errors.py
@@ -6,3 +6,6 @@ class MessageRejectedError(NylasError):
 
 class FileUploadError(NylasError):
     pass
+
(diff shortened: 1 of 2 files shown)


jetzhou:
@jhatch28 thanks for merging! Could you release a new version in PyPI? Thanks!

josefdaly:
Hey, when do you plan on releasing this?

PR #106, opened 2019-02-22
Scopes

singingwolfboy:
Implement [selective sync](https://docs.nylas.com/docs/how-to-use-selective-sync). Note that technically, this is _not_ a backwards-compatible change, because the SDK was previously setting `scope=email` on all requests. Now, the default is `scopes=email,calendar,contacts`.

Ruby SDK: https://github.com/nylas/nylas-ruby/pull/206


nylas/client/client.py
```diff
@@ -62,6 +63,8 @@ def __init__(self, app_id=environ.get('NYLAS_APP_ID'),
         self.app_secret = app_secret
         self.app_id = app_id
 
+        self.scopes = scopes or ['email']
```
emorikawa: Don't we want to default to all scopes?
```suggestion
        self.scopes = scopes or ['email', 'calendar', 'contacts]
```

tests/test_client.py
```diff
@@ -80,6 +80,35 @@ def test_client_authentication_url(api_client, api_url):
...
+    expected = (
+        URLObject(api_url)
+        .with_path("/oauth/authorize")
+        .set_query_params([
+            ('login_hint', ''),
+            ('state', ''),
+            ('redirect_uri', '/redirect'),
+            ('response_type', 'code'),
+            ('client_id', 'None'),
+            ('scopes', 'email,calendars,contacts'),
```
emorikawa: It might also be good to handle the no-scopes case and maybe a 1-scope case (like only calendar).

Also I think it's `calendar` (singular) not `calendars`

nylas/client/client.py
singingwolfboy: Depends on which we care about more: backwards compatibility, or consistency with the other SDKs. Right now, the SDK is sending the `scope` (singular) parameter set to "email", so the most backwards-compatible option is to continue sending that by default.

However, I'm inclined to agree that switching the default to all scopes is probably a better change. It's probably what our users will expect, and it's consistent with the other SDKs as well.

tests/test_client.py
singingwolfboy: You're right, it's `calendar` -- fixed.

Do we want to allow the developer to set no scopes? Right now, the codebase will treat that the same way as though the developer didn't pass a `scopes` argument at all, and so it will use the default of all scopes.

CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 nylas-python Changelog
 ======================
 
+Unreleased
```
jhatch28:
```suggestion
Unreleased (dev)
```

nylas/client/client.py
```diff
@@ -104,11 +104,16 @@ def access_token(self, value):
             if 'Authorization' in self.session.headers:
                 del self.session.headers['Authorization']
 
-    def authentication_url(self, redirect_uri, login_hint='', state=''):
+    def authentication_url(self, redirect_uri, login_hint='', state='', scopes=None):
```
jhatch28:
```suggestion
    def authentication_url(self, redirect_uri, login_hint='', state='', scopes=('email', 'calendar', 'contacts')):
```

jhatch28: nitpick: I think it might be better to put the default in the parameters so people know what it defaults to reading the code

singingwolfboy:
@jhatch28: could you please re-review this? I took your suggestion of moving the default scopes to the function parameters, but then I wasn't sure what should happen if someone explicitly passed `scopes=None` to this function. I figured that if someone bothered to do that, they probably didn't want the `scopes` query parameter to be present in the URL at all, so I changed the code to account for that, and I added another test. Does that seem reasonable to you?

jhatch28:
@singingwolfboy yeah that seems perfectly reasonable to me, thanks for getting those changes in! 

I see there's some conflicts, now. I'll take a look and see if I'm capable of resolving them

singingwolfboy:
Ah, no worries. I'll rebase, update the PR, and merge once the tests pass. Thanks!

singingwolfboy:
Oops, you were faster than me! Thanks for handling it. :)

jhatch28:
@singingwolfboy are you able to rebase this branch locally so we can merge it?

singingwolfboy:
@jhatch28 isn't it ready to merge now?

PR #108, opened 2019-02-28
Format code with `black`

singingwolfboy:
This pull request replaces [pylint](https://www.pylint.org/) with [Black](https://github.com/ambv/black), a more modern code formatting system.


smrutherford:
Oh just realized you requested me as a reviewer on this 😅. I don't know much about Black so I would defer to Jonny on this.

jhatch28:
I'm definitely not the go-to at Nylas on auto-formatters, so I'm checking with the rest of the team on it.

Biggest red flag to me is the insistence of it using `"` instead of `'`

singingwolfboy:
Here is the [explanation for Black's preference for double-quoted strings](https://github.com/ambv/black#strings):

> Why settle on double quotes? They anticipate apostrophes in English text. They match the docstring standard described in [PEP 257](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring). An empty string in double quotes (`""`) is impossible to confuse with a one double-quote regardless of fonts and syntax highlighting used. On top of this, double quotes for strings are consistent with C which Python interacts a lot with.
>
> On certain keyboard layouts like US English, typing single quotes is a bit easier than double quotes. The latter requires use of the Shift key. My recommendation here is to keep using whatever is faster to type and let Black handle the transformation.

jhatch28:
Had someone opt for Black, so we're good to go

PR #111, opened 2019-02-28
add revoke-all endpoint

smrutherford:
Adding support for `revoke-all` endpoint: https://docs.nylas.com/v2.0/reference#revoke-all


tests/test_client.py
```diff
@@ -1,6 +1,6 @@
 import re
 import json
-from six.moves.urllib.parse import parse_qs
+from six.moves.urllib.parse import parse_qs  # noqa
```
smrutherford: This was breaking the Travis CI build. Below is the Travis CI error:

________________________ [pylint] tests/test_client.py _________________________
W:  3, 0: Relative import 'six.moves.urllib.parse', should be 'six.moves' (relative-import)

The error seems to be an open pylint issue. See https://github.com/PyCQA/pylint/issues/2180

README.md
```diff
@@ -83,6 +83,17 @@ use the `revoke_token` method on APIClient
   client.revoke_token()
 ```
 
+**Revoke either all tokens or all but one token**
```
pfista:
```suggestion
**Revoke all tokens or all but one token**
```

tests/test_client.py
singingwolfboy: I suggested switching from Pylint to Black in #108. That would resolve the problem, as well 😄

nylas/client/client.py
```diff
@@ -142,6 +143,22 @@ def revoke_token(self):
         self.auth_token = None
         self.access_token = None
 
+    def revoke_all_tokens(self, keep_access_token=None):
+        account = self.account
+        client_id = self.app_id
```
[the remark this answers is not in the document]
smrutherford: yep, good point! i've updated it 🙂

README.md
smrutherford: done!

tests/test_client.py
smrutherford: haha good to know! Maybe one day lol

PR #114, opened 2019-03-12
Adding /ip_addresses endpoint and test coverage

amanda103:
adding new /ip_addresses endpoint and test coverage

### DIFF
--- a/nylas/client/client.py
+++ b/nylas/client/client.py
@@ -77,6 +77,7 @@ class APIClient(json.JSONEncoder):
         self.revoke_all_url = (
             api_server + "/a/{client_id}/accounts/{account_id}/revoke-all"
         )
+        self.ip_addresses_url = api_server + "/a/{client_id}/ip_addresses"
 
         self.app_secret = app_secret
         self.app_id = app_id
@@ -197,6 +198,12 @@ class APIClient(json.JSONEncoder):
         if keep_access_token != self.access_token:
             self.auth_token = None
             self.access_token = None
+
+    def ip_addresses(self):
+        ip_addresses_url = self.ip_addresses_url.format(client_id=self.app_id)
+        resp = self.admin_session.get(ip_addresses_url)
+        _validate(resp).json()
+        return resp.json()
 
     @property
     def account(self):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1175,3 +1175,27 @@ def mock_revoke_all_tokens(mocked_responses, api_url, account_id, app_id):
         status=200,
         body=json.dumps({"success": True}),
     )
+
+
+@pytest.fixture
+def mock_ip_addresses(mocked_responses, api_url, app_id):
+    ip_addresses_url = "{base}/a/{app_id}/ip_addresses".format(
+        base=api_url, app_id=app_id
+    )
+    mocked_responses.add(
+        responses.GET,
+        ip_addresses_url,
+        content_type="application/json",
+        status=200,
+        body=json.dumps(
+            {
+                "ip_addresses": [
+                    "39.45.235.23",
+                    "23.10.341.123",
+                    "12.56.256.654",
+                    "67.20.987.231",
+                ],
+                "updated_at": 1552072984,
+            }
+        ),
+    )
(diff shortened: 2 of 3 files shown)


tests/test_accounts.py
```diff
@@ -22,6 +22,12 @@ def test_account_json(api_client, monkeypatch):
     assert isinstance(result, dict)
 
 
+@pytest.mark.usefixtures("mock_ip_addresses")
+def test_ip_addresses(api_client_with_app_id):
+    result = api_client_with_app_id.ip_addresses()
+    assert isinstance(result, dict)
```
annielcook: Maybe also check that the dict has keys "ip_addresses" and "updated_at"

tests/conftest.py
```diff
@@ -1175,3 +1175,20 @@ def mock_revoke_all_tokens(mocked_responses, api_url, account_id, app_id):
...
+    ip_addresses_url = "{base}/a/{app_id}/ip_addresses".format(
+        base=api_url, app_id=app_id
+    )
+    mocked_responses.add(
+        responses.GET,
+        ip_addresses_url,
+        content_type="application/json",
+        status=200,
+        body = json.dumps({
+                        "ip_addresses": ["lots-of-ips"],
```
singingwolfboy: Please make the content of this mock actually correct, at least in form. Does this endpoint return IPv4 addresses? If so, I'd suggest using values like `["1.2.3.4", "192.168.0.0", "255.12.147.89"]`, etc. Does this endpoint return IPv6 addresses? If so, I'd suggest using values like `["2001:0db8:0000:0042:0000:8a2e:0370:7334", "2001:0db8:ac10:fe01::", "AbCd:ABCD:abcd::", "2001:0db8:0000:0000:0000:ff00:0042:8329", "::1"]`. Does this endpoint return both? Then mix them up in the mock content.

nylas/client/client.py
```diff
@@ -198,6 +199,15 @@ def revoke_all_tokens(self, keep_access_token=None):
             self.auth_token = None
             self.access_token = None
 
+    def ip_addresses(self):
+        ip_addresses_url = self.ip_addresses_url.format(
+            client_id=self.app_id)
+        headers = {"Content-Type": "application/json"}
+        headers.update(self.admin_session.headers)
```
singingwolfboy: Is there a reason why _this_ endpoint needs the `Content-Type` header, but others do not? We should make it consistent.

nylas/client/client.py
```diff
@@ -198,6 +199,15 @@ def revoke_all_tokens(self, keep_access_token=None):
...
+        resp = self.admin_session.get(ip_addresses_url, headers=headers)
```
singingwolfboy: Why no `_validate()`?

amanda103: will add!

nylas/client/client.py
amanda103: nope - was just following the format I saw for other endpoints. Should I remove it?

tests/conftest.py
amanda103: IPv4! I'll make that change.

nylas/client/client.py
singingwolfboy: To be honest, _all_ of the admin endpoints should probably set this header -- but let's not do that in this pull request. Consistency makes code easier to reason about, which is why we want to be consistent here.

I recommend removing the `Content-Type` header from this method, so that it's consistent with all the other methods that use admin endpoints. If you want to go the extra mile, I suggest making a _new_ pull request that sets the `Content-Type` header for _all_ the admin endpoints, and we can discuss it in _that_ pull request. That way, it's consistent and more correct. Reasonable?

tests/conftest.py
singingwolfboy: As an aside, does Nylas have any plans to support IPv6 for this feature? Is it possible that this endpoint could return IPv6 values in the future? If so, maybe we should plan for that somehow...

nylas/client/client.py
amanda103: Sounds good. I'll remove it from this endpoint.

tests/conftest.py
amanda103: To my knowledge no, but I'll check on that. For now I think having IPv4 in the fixture should be sufficient for this PR

amanda103:
Was having a little trouble with the git flow to work on this commit again, which is why it might not look streamlined. Please let me know if you have any other feedback!

PR #115, opened 2019-03-14
Update message model to include Reply-To data

torre76:
Hello,

When working with python client, I needed the Reply-To field (which is exposed on raw data).
It is missing on Python Client, although.

I extracted it in the pull request below (which I am actually using).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,4 @@ local/
 
 pip-selfcheck.json
 tests/output
+.idea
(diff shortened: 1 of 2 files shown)


torre76:
Sorry @firefly21, 

I had no time to produce the test you required. We used these changes directly from my repository because we needed that information.

I will eventually create the tests, but not very soon.

Regards

Issue #86, opened 2017-08-07
Credentials naming should be consistent


singingwolfboy:
Most of the SDK refers to the Nylas OAuth credentials as "app ID" and "app secret". However, we apparently want to start shifting over to the more standard names: "client ID" and "client secret". Whichever we pick, we should be consistent.

Current usage of "app ID" and "app secret":
* Constructor arguments for `APIClient`.
  `client = APIClient(app_id="foo", app_secret="bar")`
* Properties of `APIClient`.
  `client.app_id == "foo"; client.app_secret == "bar"`
* Environment variables for autoloading credentials: `NYLAS_APP_ID` and `NYLAS_APP_SECRET`
* pytest fixture in `conftest.py`: `app_id`
* Existing Nylas developer console refers to the credentials as "APP ID" and "APP SECRET"
* The Nylas API error message refers to "App ID" when this value is missing: ```mywebsite.com/login/nylas/authorized?reason=Sorry%2C+you+must+provide+a+valid+`client_id`.+The+`client_id`+request+parameter+should+be+the+App+ID+of+your+application+in+the+Nylas+Developer+Portal.&error=access_denied```

Current usage of "client ID" and "client secret":
* Configuration JSON files for new examples
* Configuration for [Flask-Dance](https://flask-dance.readthedocs.io/en/latest/providers.html#module-flask_dance.contrib.nylas)
* Specifically requested by @pfista: https://github.com/nylas/nylas-python/pull/74#discussion_r130416841

jeffhou:
I'm just testing out the API, and the first thing I found was this discrepancy that left me a little confused. After registration, https://dashboard.nylas.com/applications displays an `APP ID` for the default application. The first code sample in the docs (https://docs.nylas.com/reference#oauth) then asks for a `client_id` to create the valid API call.

A possible pain point especially for onboarding new users.

mrashed-dev:
The inconsistency has been fixed and now we are using `client_id` and `client_secret` for credentials as used in other Nylas components like the API, Dashboard, and other SDKs. Now, if you were previously initializing `APIClient` by passing in the credentials as a variable assignment, you are required to update your code to reflect the new credential variable names:
```
client = APIClient(
        client_id="CLIENT_ID",
        client_secret="CLIENT_SECRET",
    )
```

If you were initializing it by just passing in the values in order like so:
```
client = APIClient(
        "CLIENT_ID",
        "CLIENT_SECRET",
    )
```
then no change is required on your part.

PR #116, opened 2019-04-01
change to client_id and client_secret naming

smrutherford:
Addresses open issue #86 Credentials naming should be consistent - adds deprecation warning for September 2019 and migrates to `client_id` and `client_secret` naming.

### DIFF
--- a/nylas/client/client.py
+++ b/nylas/client/client.py
@@ -3,6 +3,7 @@ import sys
 from os import environ
 from base64 import b64encode
 import json
+from warnings import warn
 
 import requests
 from urlobject import URLObject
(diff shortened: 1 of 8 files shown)


nylas/client/client.py
```diff
@@ -89,30 +91,54 @@ def __init__(
...
-                "X-Nylas-Client-Id": self.app_id,
+                "X-Nylas-Client-Id": self.client_id,
                 "User-Agent": version_header,
             }
         super(APIClient, self).__init__()
 
+    @property
+    def app_id(self):
+        warn(DeprecationWarning('"app_id" naming convention will be deprecated Sept 2019. '
+                                'Use "client_id" instead.'))
```
Gerdie: np: let's say `app_id` will be deprecated in version 5.0.0. Use `client_id` instead.

smrutherford: Good idea!

smrutherford:
> looking good! can we add a test to make sure the warnings are emitting as expected, and a backwards compatibility test to make sure initializing the client with the `app_id` and `client_id` works as expected?

yes! Added a test. However, the warnings don't work as we thought (well, at least what I thought). The warnings only emit if the `-Wd` flag is used when opening python. Reason behind this being that warnings shouldn't emit for the user, but for the developer. So the warnings are shown in the code. Also, I changed from `DeprecationWarning` to `PendingDeprecationWarning` since it's not fully deprecated yet.

PR #122, opened 2019-08-08
Add support for `/token_info` endpoint

singingwolfboy:
https://changelog.nylas.com/new-token-info-endpoint-110263

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ nylas-python Changelog
 
 Unreleased (dev)
 ----------------
-nothing yet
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -3,7 +3,7 @@ nylas-python Changelog
 
 Unreleased (dev)
 ----------------
-nothing yet
+* Add support for `/token_info` endpoint.
```
jhatch28: Would be nice to explain to people what this endpoint is for. Something like, 

> Add support for `/token_info` endpoint, which allows you to query the available scopes and validity of a given access token for an account

PR #133, opened 2020-01-16
Add `provider` to Account model

singingwolfboy:


singingwolfboy:
@firefly21 we haven't tested any of the attrs on any of the models -- it's not clear to me that it's worth doing so. If it is, that effort probably belongs in a different pull request, since we should probably write tests for _all_ the attrs.

PR #135, opened 2020-01-20
Add recurrence rules for events (dateutil)

singingwolfboy:
This pull request adds recurrence rules for creating events. If the [`dateutil` module](https://dateutil.readthedocs.io/en/stable/) is installed, the SDK will use `dateutil.rrule.rrule` objects to represent recurrence rules; otherwise, they will be plain text strings.

I haven't done much testing with this yet -- I just wrote an automated test as a sanity check, not to be exhaustive. Definitely interested in feedback on the direction here -- should we make `dateutil` a required dependency? Should we handle this differently somehow?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,14 +3,10 @@ language: python
 python:
   - "2.7"
   - "3.6"
-env:
-  - DATEUTIL=1
-  - DATEUTIL=0
 install:
   - "travis_retry pip install -U pip pytest pylint"
   - "travis_retry pip install .[test]"
   - "travis_retry pip install codecov"
-  - "if [[ $DATEUTIL == 1 ]]; then travis_retry pip install python-dateutil; fi"
 script: "python setup.py test"
 after_success:
   - "codecov"
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,14 @@ language: python
 python:
   - "2.7"
   - "3.6"
+env:
+  - DATEUTIL=1
+  - DATEUTIL=0
 install:
   - "travis_retry pip install -U pip pytest pylint"
   - "travis_retry pip install .[test]"
   - "travis_retry pip install codecov"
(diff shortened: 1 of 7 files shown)


singingwolfboy:
@pengfeiye instead of just approving this pull request, could you please address some of the questions I mentioned in the text? This is meant to be a discussion, not a yes/no question.

pengfeiye:
> @pengfeiye instead of just approving this pull request, could you please address some of the questions I mentioned in the text? This is meant to be a discussion, not a yes/no question.

ah, sorry, didn't realize that. 
Well, I think totally makes sense to make `dateutil` as a required dependency in this case so the new clients that use the SDK can use the recurring event feature properly via the SDK. I would like to know what your concerns are for introducing a new required dependency on the SDK.

singingwolfboy:
My concern is that it's only used for a very small part of our API, and I thought that dateutil depended on many other libraries. But after looking at the dateutil source code, it looks like the only library it depends on is `six`, which is already in the dependency list.

Do you think it would be better to make dateutil always required?

pengfeiye:
I think it would be better if we just make `dateutil` as a required dependency in this case. Although it is only a small portion of our code the recurring event has always been a very important aspect of our calendar/event feature to most of our clients that uses the feature. And the fact that it would not require any other extra modules on the client side I think this would be the best approach.

singingwolfboy:
After testing this against the live Nylas API, I discovered that the format that the API accepts for recurrence rules is not the same format as `dateutil` generates when serializing to a string. (`rrule` objects in `dateutil` are required to specify a starting date, and that information is included in the string representation. However, the Nylas API wants you to include the starting date in the when object, and throws an error if it's present in the `rrule` string representation.)

I also realized that it's _already possible_ to pass recurrence rules in the SDK, as long as you pass strings that are formatted correctly. As a result, we can't require the use of `dateutil`, since it would cause backwards compatibility issues.

I'm just going to close this pull request, to avoid making the situation even more complicated than it already is.

PR #137, opened 2020-01-30
Check message ID before trying to RSVP to an event

singingwolfboy:
The Nylas API does not allow RSVPing to events without an associated message. Since we know that in advance, the SDK can raise an error before even sending the HTTP request. This not only saves bandwidth, but makes the error clearer for the developer.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1114,6 +1114,7 @@ def mock_events(mocked_responses, api_url):
         [
             {
                 "id": "1234abcd5678",
+                "message_id": "evh5uy0shhpm5d0le89goor17",
                 "title": "Pool party",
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -1130,7 +1131,13 @@ def mock_events(mocked_responses, api_url):
                         "status": "no",
                     },
                 ],
-            }
+            },
+            {
+                "id": "9876543cba",
+                "message_id": None,
```
firefly21: Would this have a None value, or would the `message_id` key be omitted entirely?

singingwolfboy: It would have a None value.

PR #139, opened 2020-03-04
Add room resources

singingwolfboy:
still need to manually test this...


BenLloydPearson:
I tested this on an account with the appropriate permissions and can confirm that I'm able to access room resources

PR #142, opened 2020-03-10
Fix limit param

singingwolfboy:
Previously, if you specified a limit of 50 or more, you would receive *ALL* the resources available on the server. This changes the SDK to properly respect the limit provided.

### DIFF
--- a/nylas/client/restful_model_collection.py
+++ b/nylas/client/restful_model_collection.py
@@ -22,19 +22,26 @@ class RestfulModelCollection(object):
         return self.values()
 
     def values(self):
+        limit = self.filters.get("limit")
         offset = self.filters["offset"]
+        fetched = 0
         while True:
-            models = self._get_model_collection(offset, CHUNK_SIZE)
+            if limit:
+                if fetched >= limit:
+                    break
+
+                req_limit = min(CHUNK_SIZE, limit - fetched)
+            else:
+                req_limit = CHUNK_SIZE
+
+            models = self._get_model_collection(offset + fetched, req_limit)
             if not models:
                 break
 
             for model in models:
                 yield model
 
-            if len(models) < CHUNK_SIZE:
-                break
-
-            offset += len(models)
+            fetched += len(models)
 
     def first(self):
         results = self._get_model_collection(0, 1)
--- a/nylas/client/restful_model_collection.py
+++ b/nylas/client/restful_model_collection.py
@@ -25,6 +25,14 @@ class RestfulModelCollection(object):
         limit = self.filters.get("limit")
         offset = self.filters["offset"]
         fetched = 0
+        # Currently, the Nylas API handles pagination poorly: API responses do not expose
+        # any information about pagination, so the client does not know whether there is
+        # another page of data or not. For example, if the client sends an API request
+        # without a limit specified, and the response contains 100 items, how can it tell
+        # if there are 100 items in total, or if there more items to fetch on the next page?
+        # It can't! The only way to know is to ask for the next page (by repeating the API
+        # request with `offset=100`), and see if you get more items or not.
+        # If it does not receive more items, it can assume that it has retrieved all the data.
         while True:
             if limit:
                 if fetched >= limit:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,6 +12,24 @@ from nylas import APIClient
 
 # pylint: disable=redefined-outer-name,too-many-lines
 
+#### HANDLING PAGINATION ####
+# Currently, the Nylas API handles pagination poorly: API responses do not expose
+# any information about pagination, so the client does not know whether there is
+# another page of data or not. For example, if the client sends an API request
+# without a limit specified, and the response contains 100 items, how can it tell
+# if there are 100 items in total, or if there more items to fetch on the next page?
+# It can't! The only way to know is to ask for the next page (by repeating the API
+# request with `offset=100`), and see if you get more items or not.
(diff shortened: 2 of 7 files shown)


tests/conftest.py
```diff
@@ -132,29 +133,34 @@ def mock_account(mocked_responses, api_url, account_id):
...
+            "linked_at": 1500920299,
+            "sync_state": "running",
+        }
+    ]
+
+    def list_callback(request):
+        url = URLObject(request.url)
+        offset = int(url.query_dict.get("offset") or 0)
+        if offset:
+            return (200, {}, json.dumps([]))
```
benjaminwhtan: For my clarification, are we returning an empty array if there's an `offset` because we aren't testing with non-zero `offset`s?

nylas/client/restful_model_collection.py
```diff
@@ -22,19 +22,26 @@ def __iter__(self):
         return self.values()
 
     def values(self):
+        limit = self.filters.get("limit")
```
benjaminwhtan: nitpick: what do you think about using bracket notation or updating the following line to use `.get` too to keep consistent?

nylas/client/restful_model_collection.py
```diff
@@ -22,19 +22,26 @@ def __iter__(self):
...
         offset = self.filters["offset"]
+        fetched = 0
         while True:
-            models = self._get_model_collection(offset, CHUNK_SIZE)
+            if limit:
+                if fetched >= limit:
+                    break
+
+                req_limit = min(CHUNK_SIZE, limit - fetched)
```
benjaminwhtan: looks like we want to have a hard limit of `CHUNK_SIZE` that we send to `_get_model_collection` at a time. So if we have a limit > `CHUNK_SIZE` we loop and grab models in batches of `CHUNK_SIZE` or whatever we need remaining until we've gotten `limit` number of models. 

I must be missing something - how are we breaking out of the loop if the limit isn't set? Will something cause `self._get_model_collection(offset + fetched, req_limit)` to return falsey the second time around?

nylas/client/restful_model_collection.py
singingwolfboy: I'd rather not do that, actually. `.get()` can return a value or `None`, while indexing (using bracket notation) will _always_ return a value (or raise an exception). Since we do `filters.setdefault("offset", offset)` on line 15, there is guaranteed to be a value for `filters["offset"]`, but there is _not_ guaranteed to be a value for `filters["limit"]`. That difference is important, and the difference between `.get()` and indexing is designed to highlight it.

Alternatively, we could make that difference go away, by removing the `.setdefault()` call on line 15. That might even be a better option, because it preserves the fact that the model collection was created without a defined limit. However, that would be a larger change, and I feel like it doesn't belong in this pull request.

nylas/client/restful_model_collection.py
singingwolfboy: > So if we have a limit > `CHUNK_SIZE` we loop and grab models in batches of `CHUNK_SIZE` or whatever we need remaining until we've gotten `limit` number of models.

That's right. It's a fairly typical pagination system, I believe.

> how are we breaking out of the loop if the limit isn't set?

Yep, I wrestled with this for awhile while writing this pull request. The basic problem is that our API doesn't expose any information about pagination, so the client has to guess. If you send an API request without a `limit` specified, and the response contains 100 items, how can you tell if there are 100 items in total, or if there more items to fetch on the next page? You can't! The only way to know is to _ask_ for the next page (by repeating the API request with `offset=100`), and see if you get more items or not. That's how we break out of the loop if the limit isn't set -- you'll notice that a few lines down, there's a check for `if not models: break`. If we make an API request that returns nothing, then break out of the loop.

Clients could avoid this pointless API request if we exposed pagination information, [like how GitHub does in their API](https://developer.github.com/v3/guides/traversing-with-pagination/). It would be nice if we could be explicit about when the client has reached the last page of data, rather than forcing clients to keep requesting until there's no more data to send. But that would require a change to the API, which is an entirely different codebase.

tests/conftest.py
singingwolfboy: See my other comment about pagination: the only way for the client to tell if it's reached the end of the data set is to keep requesting until it receives an empty list in the response. Since almost all of these mocks return only a single page of data, I simplified the mock so that if it receives a non-zero offset, it will return an empty response. So the first API request will not have an offset (or will have `offset=0`), and will return the mock data. The second API request will have `offset=50`, and will return an empty list. This allows the test to finish, rather than continually requesting pages and pages of data where every page is the same set of mock data, forever.

nylas/client/restful_model_collection.py
benjaminwhtan: Ah I see.  It would indeed be nice to return that info!
This makes sense to me now, but what do you think about adding a comment to clarify this behavior to someone coming in with less context?

nylas/client/restful_model_collection.py
benjaminwhtan: `__init__` defaults `offset` to 0 so I think keeping `filters.setdefault("offset", offset)` on line 15 makes sense. It might make even more sense to remove the `offset` param completely from `init` and have it passed in as a part of filter. That way, we could remove the `set_default`. In any case, that doesn't belong in this PR either.

This is fine to me as is!

tests/conftest.py
benjaminwhtan: Ah this makes sense. Going off my other suggestion, what do you think about leaving a short comment to clarify the behavior to someone coming in with less context?

PR #152, opened 2021-05-18
[58500] Python SDK support for metadata for events

mrashed-dev:
# Description
The Nylas Calendar API has a new beta feature of adding a new [metadata field to calendar events](https://www.nylas.com/blog/event-metadata). This PR makes the feature available via the Nylas Python SDK.

# Usage
To filter using either `metadata_key` or `metadata_value` you can pass in a string:
`events = nylas.events.where(metadata_key='hello').all()`
or, you can pass in multiple strings as an array:
`events = nylas.events.where(metadata_value=['value1', 'value2']).all()`

To filter on `metadata_pair` you can pass in a `dict` of key-value pairs to use in the query:
`events = nylas.events.where(metadata_pair={'hello': 'world'}).all()`

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ nylas-python Changelog
 
 Unreleased (dev)
 ----------------
-nothing yet
+* Add `metadata` field in the Event model to support new event metadata feature
+* Add filtering support for `metadata_pair`
 
 v4.12.1
 -------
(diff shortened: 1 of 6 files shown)


philrenaud (commented):
You'll need to run python Black locally to format things (it may just be upset about your import order; unsure)

And ideally you could add some tests to https://github.com/nylas/nylas-python/blob/main/tests/test_events.py here as well

mrashed-dev:
> You'll need to run python Black locally to format things (it may just be upset about your import order; unsure)
> 
> And ideally you could add some tests to https://github.com/nylas/nylas-python/blob/main/tests/test_events.py here as well

Black was re-ran and tests were added!

philrenaud (approved):

PR #153, opened 2021-06-10
[62077] Fix adding a tracking object to an existing `draft`

mrashed-dev:
# Description
The Python SDK creates a special payload specifically for sending an already-existing `draft` object that extracts the draft ID and the draft's version and sends that as a payload. However, it does not check if a tracking object exists, and thus if a user were to save a draft, and come add a tracking object to it (or if the user saves the draft with the tracking object first) and tries to send it after the save, the SDK never sends the tracking object in the payload. This PR adds a check to add a tracking object if present in the `draft`.

# License

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -705,35 +705,10 @@ def mock_draft_sent_response(mocked_responses, api_url):
 
 @pytest.fixture
 def mock_draft_send_unsaved_response(mocked_responses, api_url):
-    body = {
-        "id": None,
-        "bcc": [],
-        "body": "Our latest sale!",
-        "cc": [],
-        "date": 1438684486,
-        "events": [],
-        "files": [],
-        "folder": None,
-        "from": [{"email": "benb@nylas.com"}],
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -703,6 +703,46 @@ def callback(request):
...
+        "unread": False,
+        "version": 0,
+    }
+
+    values = [(400, {}, "Couldn't send email"), (200, {}, json.dumps(body))]
+
+    def callback(request):
+        payload = json.loads(request.body)
+        payload["draft_id"] = "2h111aefv8pzwzfykrn7hercj"
+        return values.pop()
```
jseller, line 736: what happens with the payload? its parsed but isn't used?

mrashed-dev: Great call out. I copied over the mock from another existing mock but I agree it doesn't make too much sense nor does it need to be so complicated. I simplified the mock to fit what is actually needed. For the specific test case I used this for I don't need the payload but for the sake of mocking I still return a payload. I think this SDK needs a refactor of the test cases at some point.

jseller (approved):

PR #154, opened 2021-06-10
[62132] Fix issue when converting offset-aware `datetime` objects to `timestamp`

mrashed-dev:
# Description
Wherever we accept a date as a filtering parameter, we usually convert it from a `datetime` object to a `timestamp` object. However the way we convert it does not take into consideration if the `datetime` is offset-aware or not (timezone), and thus would throw an error when it's attempted. Now we check if the `datetime` is context-aware, and if it is, we convert it properly to being an offset-naive object before converting it to a timestamp.

# License

### DIFF
--- a/nylas/utils.py
+++ b/nylas/utils.py
@@ -1,5 +1,5 @@
 from __future__ import division
-from datetime import datetime
+from datetime import datetime, timedelta
 
 
 def timestamp_from_dt(dt, epoch=datetime(1970, 1, 1)):
@@ -13,7 +13,7 @@ def timestamp_from_dt(dt, epoch=datetime(1970, 1, 1)):
 
     delta = dt - epoch
 
-    return delta.seconds + delta.days * 86400
+    return int(delta.total_seconds() / timedelta(seconds=1).total_seconds())
 
 
 def convert_datetimes_to_timestamps(data, datetime_attrs):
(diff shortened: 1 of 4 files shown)


nylas/utils.py
```diff
@@ -7,8 +7,12 @@ def timestamp_from_dt(dt, epoch=datetime(1970, 1, 1)):
...
     https://stackoverflow.com/a/8778548/141395
     """
+    # For offset-aware datetime objects, convert them first before performing delta
+    if dt.tzinfo is not None and dt.utcoffset() is not None:
+        dt = dt.replace(tzinfo=None) - dt.utcoffset()
+
     delta = dt - epoch
-    # return delta.total_seconds()
+
     return delta.seconds + delta.days * 86400
```
jseller, line 17: maybe a good idea to use timedelta from the datetime package for adding or subtracting time values from datetime. 

Also, datetime has a timestamp() method, so that could be just: return delta.timestamp()

mrashed-dev: Thanks for pointing my attention to `timedelta` it helps clean the code a bit! Regarding `datetime.timestamp()` I initially used that as the solution, but unfortunately it's only available from Python 3.3 and up, and we still support Python 2.7 I believe so I had to scrap that idea.

jseller: ok, this is good for py2, that timestamp() is in py3. datetime in python isn't the best package

jseller (approved):
lgtm

nylas/utils.py
mrashed-dev: So I've heard :/

PR #155, opened 2021-06-14
[47453] Fix `limit` value in filter not being used when making `.all()` call

mrashed-dev:
# Description
There's a bug where if the user adds a `limit` value as a filter before calling `.all()`, their value will be discarded (e.g. `nylas.threads.where(limit=50).all()`. Now we properly check if the filter exists and apply the setting before making the request to the API, resulting in the expected behaviour. 

# License


yusra-ahmed (approved):
lgtm

PR #157, opened 2021-06-14
[62344] Added `Nylas-API-Version` header support

mrashed-dev:
# Description
Now we specify the `Nylas-API-Version` header and send it with every request made to the Nylas API. With this change:

* There is a new `SUPPORTED_API_VERSION` value that acts as the default API version
* Users are able to override this value and use an API version of their choice to specify a `api_version` when initializing a new `APIClient`

# Usage
The usage does not change if you are fine with using the latest supported API. However, if you want to use a different version you may create a new instance of `APIClient` like so:
```
nylas = APIClient(
    app_id=CLIENT_ID,
    app_secret=CLIENT_SECRET,
    access_token=ACCESS_TOKEN,
    api_version='2.1'
)
```

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Unreleased (dev)
 ----------------
 * Add `metadata` field in the Event model to support new event metadata feature
 * Add filtering support for `metadata_pair`
+* Added `Nylas-API-Version` header support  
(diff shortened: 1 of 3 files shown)


yusra-ahmed (approved):
lgtm

PR #158, opened 2021-06-15
[62389] Remove `bumpversion` as a runtime dependency

mrashed-dev:
# Description
We don't need `bumpversion` for running the package, only for releasing, so we shouldn't include it as a required dependency. Also we switched to `subprocess` for safer execution of child processes.

# License


yusra-ahmed (approved):
lgtm

PR #159, opened 2021-06-16
[62378] Updated credential naming to be consistent with the API and other SDKs

mrashed-dev:
# Discussion
In the other Nylas components like the API and Dashboard, as well as the other SDKs, the credentials are referred to as `client_id` and `client_secret` however this SDK was referring to these credentials as `app_id` and `app_secret`. This causes some unnecessary confusion, and this change addresses this by switching all references of `app_id` and `app_secret` to `client_id` and `client_secret`.

# Usage
If you were previously initializing `APIClient` by passing in the credentials as a variable assignment, you are required to update your code to reflect the new credential variable names:
```
client = APIClient(
        client_id="CLIENT_ID",
        client_secret="CLIENT_SECRET",
    )
```

If you were initializing it by just passing in the values in order like so:
```
client = APIClient(
        "CLIENT_ID",
        "CLIENT_SECRET",
    )
```
then no change is required on your part.

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ nylas-python Changelog
 
 Unreleased (dev)
 ----------------
+* Transitioned from `app_id` and `app_secret` naming to `client_id` and `client_secret`
 * Add `metadata` field in the Event model to support new event metadata feature
 * Add filtering support for `metadata_pair`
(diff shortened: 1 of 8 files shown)


lkaileh (approved):
LGTM!

PR #163, opened 2021-07-17
[63415] Neural API support for Python SDK

mrashed-dev:
# Description
This PR brings support for the Nylas [Neural API](https://developer.nylas.com/docs/intelligence/). The Neural API exposes endpoints that perform intelligent analysis on items in your Nylas account. This change will enable users to access the features of the Nylas Neural API and provides some utility functions to help improve the user's workflow.

# Usage
To use [Sentiment Analysis](https://developer.nylas.com/docs/intelligence/sentiment-analysis/):
```python
# To perform sentiment analysis on a message, pass in the list of message ID:
message_analysis = nylas.neural.sentiment_analysis_message([MESSAGE_ID])

# To perform sentiment analysis on just text, pass in a string:
text_analysis = nylas.neural.sentiment_analysis_text("Hi, thank you so much for reaching out! We can catch up tomorrow.")
```

To use [Signature Extraction](https://developer.nylas.com/docs/intelligence/signature-extraction/):
```python
const signature = nylas.neural.extract_signature([MESSAGE_ID])

# The method also accepts two optional parameters
# parseContact, a boolean for whether Nylas should parse the contact from the signature (API defaults to true)
# options, an object of options that can be enabled for the Neural endpoint, of type NeuralMessageOptions:
options = NeuralMessageOptions(
    ignore_links=False,
    ignore_images=False,
    ignore_tables=False,
    remove_conclusion_phrases=False,
    images_as_markdowns=False
)

signature = nylas.neural.extract_signature([MESSAGE_ID], true, options)
```
and to parse the contact and convert it to the standard Nylas contact object:
```python
contact = signature[0].contacts.to_contact_object()
```

To use [Clean Conversations](https://developer.nylas.com/docs/intelligence/clean-conversations/):
```python
convo = nylas.neural.clean_conversation([MESSAGE_ID])

# You can also pass in an object of options that can be enabled for the Neural endpoint, of type NeuralMessageOptions
convo = nylas.neural.clean_conversation([MESSAGE_ID], options);
```
and to extract images from the result:
```python
convo[0].extract_images()
```

To use [Optical Character Recognition](https://developer.nylas.com/docs/intelligence/optical-charecter-recognition/):
```python
ocr = nylas.neural.ocr_request(FILE_ID)

# This endpoint also supports a second, optional parameter for an array specifying the pages that the user wants analyzed:
ocr = nylas.neural.ocr_request(FILE_ID, [2, 3])
```

To use [Categorizer](https://developer.nylas.com/docs/intelligence/categorizer/)
```python
cat = nylas.neural.categorize([MESSAGE_ID])

# You can also send a request to recategorize the message:
cat = cat[0].recategorize("conversation")
```

# License

### DIFF
--- a/nylas/client/client.py
+++ b/nylas/client/client.py
@@ -511,9 +511,10 @@ class APIClient(json.JSONEncoder):
         response = session.request(method, url, json=converted_data)
 
         result = _validate(response).json()
-        # The Neural API will always return only one item
-        # but sometimes it returns it in the form of a 'singleton' array
         if isinstance(result, list):
-            result = result[0]
+            object_list = []
+            for obj in result:
+                object_list.append(cls.create(self, **obj))
+            return object_list
 
         return cls.create(self, **result)
--- a/nylas/client/client.py
+++ b/nylas/client/client.py
@@ -27,6 +27,7 @@ from nylas.client.restful_models import (
     Label,
     Draft,
 )
+from nylas.client.neural_api_models import Neural
 from nylas.utils import convert_datetimes_to_timestamps, timestamp_from_dt
 
 DEBUG = environ.get("NYLAS_CLIENT_DEBUG")
@@ -332,6 +333,10 @@ class APIClient(json.JSONEncoder):
     def calendars(self):
         return RestfulModelCollection(Calendar, self)
 
+    @property
+    def neural(self):
+        return Neural(self)
+
     ##########################################################
     #   Private functions used by Restful Model Collection   #
     ##########################################################
@@ -493,4 +498,22 @@ class APIClient(json.JSONEncoder):
         response = session.post(url, json=converted_data)
 
         result = _validate(response).json()
+        return cls.create(self, **result)
+
+    def _request_neural_resource(self, cls, data, path=None, method="PUT"):
+        if path is None:
+            path = cls.collection_name
+        url = URLObject(self.api_server).with_path("/neural/{name}".format(name=path))
+
+        session = self._get_http_session(cls.api_root)
+
+        converted_data = convert_datetimes_to_timestamps(data, cls.datetime_attrs)
+        response = session.request(method, url, json=converted_data)
+
+        result = _validate(response).json()
+        # The Neural API will always return only one item
+        # but sometimes it returns it in the form of a 'singleton' array
+        if isinstance(result, list):
+            result = result[0]
+
         return cls.create(self, **result)
--- a/nylas/client/restful_models.py
+++ b/nylas/client/restful_models.py
@@ -21,6 +21,13 @@ def typed_dict_attr(items, attr_name=None):
 
 
 class NylasAPIObject(dict):
+def _is_subclass(cls, parent):
+    for base in cls.__bases__:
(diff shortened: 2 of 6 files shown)


nylas/client/restful_models.py
```diff
@@ -46,7 +53,12 @@ def __init__(self, cls, api):
     def create(cls, api, **kwargs):
         object_type = kwargs.get("object")
         cls_object_type = getattr(cls, "object_type", cls.__name__.lower())
-        if object_type and object_type != cls_object_type and object_type != "account":
+        if (
+            object_type
+            and object_type != cls_object_type
+            and object_type != "account"
+            and not _is_subclass(cls, object_type)
```
mrashed-dev, line 60: Many Neural objects are objects that extend an existing class (such as `Message`), but unlike `Drafts` the server payload still denotes the `object` key as `message` instead of - say 'signature'. To prevent this from breaking existing functionality, we check if the class being created is the subclass of the inbound payload's `object` value.

nylas/client/restful_models.py
```diff
@@ -20,7 +20,14 @@ def typed_dict_attr(items, attr_name=None):
...
 
-class NylasAPIObject(dict):
+def _is_subclass(cls, parent):
+    for base in cls.__bases__:
+        if base.__name__.lower() == parent:
+            return True
+    return False
+
+
+class RestfulModel(dict):
```
mrashed-dev, line 30: We do this because the Neural API series of endpoints don't support the same operations as the typical Nylas API objects. Thus this allows us to extend the minimum required functionality shared by both the typical Nylas API objects and Neural API objects.

nylas/client/client.py
```diff
@@ -494,3 +499,21 @@ def _call_resource_method(self, cls, id, method_name, data):
...
+        url = URLObject(self.api_server).with_path("/neural/{name}".format(name=path))
+
+        session = self._get_http_session(cls.api_root)
+
+        converted_data = convert_datetimes_to_timestamps(data, cls.datetime_attrs)
+        response = session.request(method, url, json=converted_data)
+
+        result = _validate(response).json()
+        # The Neural API will always return only one item
+        # but sometimes it returns it in the form of a 'singleton' array
```
egginsect, line 515: Is this the restriction of the python SDK? `conversation` and `signature` can actually return list of objects.

egginsect: If the endpoints get more than 1 message id, it will return multiple messages

mrashed-dev: Oops, I did not know that. I thought from the documentation it was always 1:1. I will update the code to reflect this.

mrashed-dev: @egginsect I have pushed a change to reflect your comments. Does the same apply to OCR? Or does OCR only support one file ID at a time? Thanks!

egginsect: Yeah it takes only 1 file_id per request

philrenaud (approved):

jhatch28 (approved):
Didn't review for potential bugs, just for the interface, which looks great

egginsect (approved):

PR #176, opened 2021-11-05
[67238] Return better error details from the API

mrashed-dev:
# Description
Previously, we would only return a vague error message thrown from the requests library, without any detail from the Nylas API. Now we return a custom error object (that extends `HTTPError`) that properly returns a message that includes the error code and the HTTP error type, as well as the error reason and error type from the Nylas API.

### Errors Before
`requests.exceptions.HTTPError: 404 Client Error: NOT FOUND for url: https://api.nylas.com/calendars/foo`

### After
`nylas.client.errors.NylasApiError: 404 NOT FOUND. Reason: Calendar foo not found. Nylas Error Type: invalid_request_error`

# Usage

# License

### DIFF
--- a/nylas/client/errors.py
+++ b/nylas/client/errors.py
@@ -29,6 +29,7 @@ class NylasApiError(HTTPError):
     Error class for Nylas API Errors
     This class provides more information to the user sent from the server, if present
     """
+
     def __init__(self, response):
         try:
             response_json = json.loads(response.text)
--- a/nylas/client/errors.py
+++ b/nylas/client/errors.py
@@ -1,3 +1,8 @@
+import json
+
+from requests import HTTPError
+
+
 class NylasError(Exception):
     pass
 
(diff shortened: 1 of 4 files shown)


nylas/client/errors.py
```diff
@@ -17,3 +22,23 @@ class UnSyncedError(NylasError):
...
+            response_json = json.loads(response.text)
+            error_message = u"%s %s. Reason: %s. Nylas Error Type: %s" % (
+                response.status_code,
+                response.reason,
+                response_json["message"],
+                response_json["type"],
+            )
+            super(NylasApiError, self).__init__(error_message, response=response)
+        except (ValueError, KeyError):
+            super(NylasApiError, self).__init__(response.text, response=response)
```
jseller, line 44: this text could be large, and why would the json parse fail here?

jseller (approved):
lgtm! nice changes

nylas/client/errors.py
mrashed-dev: The reason why JSON parsing could fail is a rare occasion where the server is down, the error returned would be an HTML or just plain text, which would result in a paring error when attempting to parse. We've had this occur with the [Ruby SDK for example](https://github.com/nylas/nylas-ruby/pull/316).

PR #180, opened 2021-12-08
[75449] Fix bug where updating an Event results in an API error

mrashed-dev:
# Description
A recent API update caused updating an event using the SDK to return an error because the SDK sends a Participant's status on an Event update call. This PR omits this field when converting the Event object to a JSON.

# License

### DIFF
--- a/nylas/client/restful_models.py
+++ b/nylas/client/restful_models.py
@@ -663,6 +663,12 @@ class Event(NylasAPIObject):
             dct["when"] = dct["when"].copy()
             dct["when"].pop("object", None)
 
+        if dct.get("participants") and isinstance(dct.get("participants"), list):
+            # The status of a participant cannot be updated and, if the key is
+            # included, it will return an error from the API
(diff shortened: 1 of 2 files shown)


nylas/client/restful_models.py
```diff
@@ -663,6 +663,12 @@ def as_json(self):
             dct["when"] = dct["when"].copy()
             dct["when"].pop("object", None)
 
+        if dct.get("participants") and isinstance(dct.get("participants"), list):
+            # The status of a participant cannot be updated and, if the key is
+            # included, it will return an error from the API
+            for participant in dct.get("participants"):
+                participant.pop("status")
```
jaykim97, line 670: does all participant instance hold "status" value? else wouldn't this cause an error?

mrashed-dev: Good catch! Pushed a commit to address this

jaykim97 (approved):
LG!

PR #181, opened 2021-12-14
[75826] Job Status support

mrashed-dev:
# Description
This PR adds support for job statuses within supported models, as well as the `/job-statuses` endpoint.

# Usage
To view all Job statuses
```ruby
job_statuses = nylas.job_statuses.all()
```

To view a specific Job status
```ruby
job_status = nylas.job_statuses.get("JOB_STATUS_ID")
```

To get a boolean value representing Job status success/failure
```ruby
job_status = nylas.job_statuses.get("JOB_STATUS_ID")
job_status.is_successful()
```

# License

### DIFF
--- a/nylas/client/restful_models.py
+++ b/nylas/client/restful_models.py
@@ -181,6 +181,7 @@ class Message(NylasAPIObject):
         "starred",
         "subject",
         "thread_id",
+        "job_status_id",
         "to",
         "unread",
         "starred",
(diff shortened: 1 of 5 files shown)


billwjo (approved):

nylas/client/restful_models.py
```diff
@@ -638,6 +645,7 @@ class Event(NylasAPIObject):
         "recurrence",
         "status",
         "master_event_id",
+        "job_status_id",
```
jseller, line 648: this job status doesn't always get returned from /event, will this parsing be lenient if it None?

mrashed-dev: @jseller That is correct, parsing will be lenient if any of the values are `None`

jseller (approved):
LG!

PR #184, opened 2021-12-23
[77068] Fix passing `kwargs` to `_update_resource`

mrashed-dev:
# Description
When passing keyword arguments to a method that utilizes `_update_resource`, such as `Event.save()`, the keyword arguments were incorrectly being bubbled up resulting in them not formatting in the URL as URL parameters.

# License


maxwell-schwartz (approved):
lgtm!

PR #186, opened 2021-12-30
[76804] Event to ICS Support

mrashed-dev:
# Description
This PR enables support for [generating an ICS file from an Event](https://developer.nylas.com/docs/api/#post/events/to-ics).

# Usage
```python
from nylas import APIClient
nylas = APIClient(
    CLIENT_ID,
    CLIENT_SECRET,
    ACCESS_TOKEN
)
example_event = nylas.events.get('{id}')

# Generate an ICS from an event
ics = example_event.generate_ics()

# You can also pass ICS Options for more configuration
ics = example_event.generate_ics(
  ical_uid="test_uuid",
  method="add",
  prodid="test_prodid"
)
```

# License

### DIFF
--- a/nylas/client/restful_models.py
+++ b/nylas/client/restful_models.py
@@ -746,7 +746,6 @@ class Event(NylasAPIObject):
             "Unexpected response from the API server. Returned 200 but no 'ics' string found."
         )
 
-
     def save(self, **kwargs):
         if (
             self.conferencing
--- a/nylas/client/restful_models.py
+++ b/nylas/client/restful_models.py
@@ -738,13 +738,13 @@ class Event(NylasAPIObject):
         if ics_options:
             payload["ics_options"] = ics_options
 
-        response = self.api._post_resource(
-            Event, None, "to-ics", payload
-        )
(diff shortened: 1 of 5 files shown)


nylas/client/restful_models.py
```diff
@@ -701,6 +701,52 @@ def rsvp(self, status, comment=None):
...
+        if ics_options:
+            payload["ics_options"] = ics_options
+
+        response = self.api._post_resource(Event, None, "to-ics", payload)
+        try:
+            return response["ics"]
+        except KeyError:
+            raise RuntimeError(
+                "Unexpected response from the API server. Returned 200 but no 'ics' string found."
+            )
```
james10424, line 748: Maybe an necessary try, could be written like
```suggestion
        if "ics" in response:
            return response["ics"]
        raise RuntimeError(
            "Unexpected response from the API server. Returned 200 but no 'ics' string found."
        )

```

mrashed-dev: @james10424 thank you! Love this change.

james10424 (approved):
LG

Issue #189, opened 2022-01-25
Expose full response dict when calling POST `/oauth/token`


yi-fan-song:
**Is your feature request related to a problem? Please describe.**

I don't ask for a user email on my end, so I need to get it from that response.

**Describe the solution you'd like**

Just another method that simply returns the response instead of the token.

```
def trade_code(self, code):
        args = {
            "client_id": self.client_id,
            "client_secret": self.client_secret,
            "grant_type": "authorization_code",
            "code": code,
        }

        headers = {
            "Content-type": "application/x-www-form-urlencoded",
            "Accept": "text/plain",
        }

        resp = requests.post(
            self.access_token_url, data=urlencode(args), headers=headers
        ).json()

        self.access_token = resp[u"access_token"]
        return resp
```

**Describe alternatives you've considered**

I'm making the call myself right now.

**Additional context**

mrashed-dev:
@yi-fan-song thank you for opening this issue! We will investigate adding this to the SDK and let you know once it's been complete & released!

mrashed-dev:
@yi-fan-song The fix has been merged into main and can be accessed via `send_authorization(code)`. It will be included in the next release coming very soon!

mrashed-dev:
@yi-fan-song [Nylas Python SDK v5.5.0](https://github.com/nylas/nylas-python/releases/tag/v5.5.0) has been released containing this new function! Below is how you can access the full payload:

```python
from nylas import APIClient
nylas = APIClient(
    CLIENT_ID,
    CLIENT_SECRET,
    ACCESS_TOKEN
)

ACCESS_TOKEN = nylas.send_authorization('{code_from_nylas}')
```

PR #190, opened 2022-01-31
[79420] Enable full payload response for exchanging the token for code

mrashed-dev:
# Description
This PR introduces a new method, `send_authorization` that will call the `/oauth/token` endpoint and returns the full payload that includes the access token within it.

# License

### DIFF
--- a/nylas/client/client.py
+++ b/nylas/client/client.py
@@ -169,7 +169,16 @@ class APIClient(json.JSONEncoder):
         url = URLObject(self.authorize_url).add_query_params(args.items())
         return str(url)
 
-    def token_for_code(self, code):
+    def send_authorization(self, code):
+        """
+        Exchanges an authorization code for an access token.
+
+        Args:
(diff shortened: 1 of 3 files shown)


nylas/client/client.py
```diff
@@ -186,7 +195,11 @@ def token_for_code(self, code):
             self.access_token_url, data=urlencode(args), headers=headers
         ).json()
 
-        self.access_token = resp[u"access_token"]
+        self.access_token = resp["access_token"]
+        return resp
+
+    def token_for_code(self, code):
+        self.send_authorization(code)
```
maxwell-schwartz, line 202: Does anything need to happen with the `resp` returned from `send_authorization`?

mrashed-dev: @maxwell-schwartz no, this was just to fulfill #189 where the user would like to have the full payload made available to them rather than us stripping the access token from the response and returning the access token.

maxwell-schwartz (approved):
lgtm!

PR #196, opened 2022-02-09
[80305] Fix `native-authentication-gmail` example app

mrashed-dev:
# Description
`client_id` and `client_secret` should be present for `make_google_blueprint` to succeed, as reported by @[GroveIV](https://github.com/GroveIV)

# License

### DIFF
--- a/examples/native-authentication-gmail/server.py
+++ b/examples/native-authentication-gmail/server.py
@@ -66,6 +66,8 @@ if cfg_needs_replacing:
 # Use Flask-Dance to automatically set up the OAuth endpoints for Google.
 # For more information, check out the documentation: http://flask-dance.rtfd.org
 google_bp = make_google_blueprint(
+    client_id=app.config["GOOGLE_OAUTH_CLIENT_ID"],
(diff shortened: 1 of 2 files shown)


examples/native-authentication-gmail/server.py
```diff
@@ -66,6 +66,8 @@
 # Use Flask-Dance to automatically set up the OAuth endpoints for Google.
 # For more information, check out the documentation: http://flask-dance.rtfd.org
 google_bp = make_google_blueprint(
+    client_id=app.config["GOOGLE_OAUTH_CLIENT_ID"],
+    client_secret=app.config["GOOGLE_OAUTH_CLIENT_SECRET"],
```
jaykim97, line 70: If the values were not set on config.json (not wrong credential) would this throw error to end user?

mrashed-dev: @jaykim97 Probably, but we tell the user to provide it via the README: https://github.com/nylas/nylas-python/tree/main/examples/native-authentication-gmail#readme

and if it isn't passed in the call will fail anyways

jaykim97 (approved):

PR #208, opened 2022-03-30
Add support for `limit` and `offset` for message/thread search

mrashed-dev:
# Description
This PR adds limit and offset as parameters for message/thread search.

# License


hafsakhalid (approved):
👍

PR #214, opened 2022-04-13
[TSDK-287] Fix Calendar availability functions not using the correct authentication method

mrashed-dev:
# Description
The Calendar availability functions were using an incorrect authentication method when requesting to the API

# License

### DIFF
--- a/nylas/client/client.py
+++ b/nylas/client/client.py
@@ -287,7 +287,7 @@ class APIClient(json.JSONEncoder):
             "start_time": start_time,
             "end_time": end_time,
         }
-        resp = self._request(HttpMethod.POST, url, json=data)
+        resp = self._request(HttpMethod.POST, url, json=data, cls=Calendar)
         _validate(resp)
         return resp.json()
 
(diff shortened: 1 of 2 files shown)


nylas/client/client.py
```diff
@@ -287,7 +287,7 @@ def free_busy(self, emails, start_at, end_at):
             "start_time": start_time,
             "end_time": end_time,
         }
-        resp = self._request(HttpMethod.POST, url, json=data)
+        resp = self._request(HttpMethod.POST, url, json=data, cls=Calendar)
```
dantonpimentel, line 290: just to clarify this for me, the class knows what scopes are required from the token? Or is it something completely different?

dantonpimentel (approved):
lgtm

nylas/client/client.py
mrashed-dev: @dantonpimentel no, this is just for the function to notify the request function to use Bearer authentication as it's part of the Calendar API

PR #221, opened 2022-05-25
Change config file to match new Flask rules

erikkai:
# Description
In Flask 2.1 onwards the method used to set up the config file is no longer available. So it will break. I replaced it with the config option where you load from a file. Then I set up config.json to be config.py so that it will work. 

# License


mrashed-dev (approved):
Looks good to me @erikkai! Just need to fix the formatting and it will be good to go 👍

PR #223, opened 2022-06-19
use http basic auth for /oauth/revoke endpoint

mkgs:
Current implementation will always return a 401. Use basic auth per the Nylas docs: https://developer.nylas.com/docs/api#post/oauth/revoke


mrashed-dev (approved):
Looks good to me, thanks @mkgs! 🚀

PR #233, opened 2022-09-20
[TSDK-483] Fix authentication for integrations

mrashed-dev:
# Description
In Python 2.7 the `'b'` prefix ahead of a string is ignored but in Python 3 it indicates a byte literal versus a string. So on Python 3 apps implementing our SDK they would encounter an error as the Nylas API is expecting a string instead of a byte string. Now we ensure that it gets decoded back into a string to ensure it works on Python 3.

# License


theblueskies (approved):

dantonpimentel (approved):
lgtm

PR #240, opened 2022-11-16
Add visibility field in Event

mrashed-dev:
# Description
Add the `visibility` to the Event class.

# License


atejada (approved):
All good 👍

PR #247, opened 2023-01-05
Only install enum34 on python34 and below.

Hooksie:
Hey folks,

Noticed earlier that nylas is our one dependency that's causing `enum34` to be added to our full dependencies, despite our codebase being well into the 3.x line. Small patch to limit `enum34` to installs on Python3.4 or less. Py34 is roughly chosen here due to the relative time of release (w.r.t. the last updates to `enum34`), and to avoid changing deps on any early-adopter py34 projects.

Ran through validation installs on 3.10.5 and 2.7.18 and checked the `pip freeze` output to verify py27 had `enum34` and py3 did not

# License

👍👍👍 

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,8 +15,11 @@ RUN_DEPENDENCIES = [
     "requests[security]>=2.4.2",
     "six>=1.4.1",
     "urlobject",
-    "enum34>=1.1.10",
 ]
+
+if sys.version_info[:2] <= (3, 4):
+    RUN_DEPENDENCIES.append("enum34>=1.1.10")
+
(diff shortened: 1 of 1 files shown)


mrashed-dev (approved):
Thanks for your contribution @Hooksie !

Hooksie:
@mrashed-dev My pleasure

I'm not super familiar with the `nylas-python` release process. Do you have a release cadence you prefer, or should I open a PR for `5.12.0` bump?

mrashed-dev:
@Hooksie we usually release by end of week so you should see one by tomorrow!

PR #250, opened 2023-01-27
Use PEP508 syntax for conditional dependency.

Hooksie:
Following up on #247, where I seem to have spaced-out and forgot the modern syntax.

Updates previous conditional dependency for `enum34` to use the modern PEP508 syntax.

# License


mrashed-dev (approved):
Love this change, thanks @Hooksie!

PR #280, opened 2023-10-04
Make Grants API available

spang:
This functionality was all implemented but not exposed on the Client object and therefore there was no way to make API calls

# License


spang:
Hrm. I have realized that this API IS available, but as a subproperty of `Client.auth`... SO, this isn't necessary. But I was pretty confused about how to get it. Might be a problem solvable with docs?

PR #284, opened 2023-10-12
Fix KeyError when building Auth URL

mrashed-dev:
# Description
We were not properly checking the existence of an entry in a dict.

# License


AaronDDM (approved):

kraju3:
@mrashed-dev This seems to fix the issue.

PR #292, opened 2023-10-17
Add Connector support

mrashed-dev:
# Description
This PR adds support for the connector endpoints.

# License

### DIFF
--- a/nylas/models/connectors.py
+++ b/nylas/models/connectors.py
@@ -81,7 +81,7 @@ class GoogleCreateConnectorRequest(BaseCreateConnectorRequest):
         scope: The Google OAuth scopes
     """
 
-    settings: NotRequired[GoogleCreateConnectorSettings]
+    settings: GoogleCreateConnectorSettings
     scope: NotRequired[List[str]]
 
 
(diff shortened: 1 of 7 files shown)


nylas/models/connectors.py
```diff
@@ -0,0 +1,162 @@
...
+    Interface representing the base Nylas connector creation request.
+
+    Attributes:
+        name (str): Custom name of the connector
+        provider (Provider): The provider type, should be Google
+        settings: The Google OAuth provider credentials and settings
+        scope: The Google OAuth scopes
+    """
+
+    settings: NotRequired[GoogleCreateConnectorSettings]
```
kraju3, line 84: Hey @mrashed-dev do you know why this is not required? I would assum the GoogleCreateConnectorSettings to be something that is required to create a Connector

mrashed-dev: @kraju3 ah you're right, in the docs for some reason settings is not required but client_id and client_secret is. Looks like a typo on the docs side. Made the change now.

kraju3 (approved):
Sweet LGTM! I will get this merged in and hopefully push out the PR for Credentials in the morning

PR #296, opened 2023-10-23
Fix get availability method call

atejada:
This PR will fix the call to the get_availability method.

# License


atejada:
Not exactly what I had in mind...closing this and opening another...

PR #297, opened 2023-10-23
Update calendars.py

atejada:
The get_availability method is not working properly. Fixed the path and the response.

# License

### DIFF
--- a/nylas/resources/calendars.py
+++ b/nylas/resources/calendars.py
@@ -114,13 +114,12 @@ class Calendars(
         )
 
     def get_availability(
-        self, identifier: str, request_body: GetAvailabilityRequest
+        self, request_body: GetAvailabilityRequest
     ) -> Response[GetAvailabilityResponse]:
         """
         Get availability for a Calendar.
 
         Args:
-            identifier: The grant ID or email account to get availability for.
             request_body: The request body to send to the API.
 
         Returns:
--- a/nylas/resources/calendars.py
+++ b/nylas/resources/calendars.py
@@ -128,11 +128,11 @@ class Calendars(
         """
         json_response = self._http_client._execute(
             method="POST",
-            path=f"/v3/grants/{identifier}/calendar/availability",
+            path=f"/v3/calendar/availability",
             request_body=request_body,
         )
 
-        return Response.from_dict(json_response, GetAvailabilityResponse)
+        return Response(json_response, GetAvailabilityResponse)
 
     def get_free_busy(
         self, identifier: str, request_body: GetFreeBusyRequest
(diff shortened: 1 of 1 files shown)


wiseaidev:
In this case, you should remove the `identifier` parameter as well.

```diff
    def get_availability(
        self,
- identifier: str,
        request_body: GetAvailabilityRequest
    ) -> Response[GetAvailabilityResponse]:
        """
        Get availability for a Calendar.
        Args:
- identifier: The grant ID or email account to get availability for.
            request_body: The request body to send to the API.
        Returns:
            Response: The availability response from the API.
        """
```

P.S. Related to #298 for the record.

atejada:
@wiseaidev You're right. I have updated my PR. Thanks 😊

nylas/resources/calendars.py
```diff
@@ -114,25 +114,24 @@ def destroy(self, identifier: str, calendar_id: str) -> DeleteResponse:
...
-            identifier: The grant ID or email account to get availability for.
             request_body: The request body to send to the API.
 
         Returns:
             Response: The availability response from the API.
         """
         json_response = self._http_client._execute(
             method="POST",
-            path=f"/v3/grants/{identifier}/calendar/availability",
+            path=f"/v3/calendar/availability",
```
mrashed-dev, line 130: should be /calendars

```suggestion
            path="/v3/calendars/availability",
```

nylas/resources/calendars.py
```diff
@@ -114,25 +114,24 @@ def destroy(self, identifier: str, calendar_id: str) -> DeleteResponse:
...
             request_body=request_body,
         )
 
-        return Response.from_dict(json_response, GetAvailabilityResponse)
+        return Response(json_response, GetAvailabilityResponse)
```
mrashed-dev, line 134: Is this tested to work? The other responses use `from_dict`

mrashed-dev (changes requested):
Thanks for opening the PR @atejada! Left a couple of notes

nylas/resources/calendars.py
atejada: I know 😅 I did it locally but forgot to rename it here 😂

nylas/resources/calendars.py
atejada: I removed it because it was generating some weird errors...once removed everything worked fine...test it several times

mrashed-dev:
Hey @atejada I just realized that Kiran's PR (#300) which was merged includes your fix. I'm going to close this PR, thanks for the contribution!

atejada:
All good 😊

PR #303, opened 2023-10-27
remove unused imports

spang:
I use pyflakes locally in my editor to catch little lint stuff like this. Seems worth keeping the repo clean.

# License


mrashed-dev (approved):

mrashed-dev:
@spang Thanks for the catch, I'll look to adding something similar to catch these before merging PRs.

PR #305, opened 2023-10-27
Add support for Read, Update, and Delete for Messages

spang:
# Description
RUD support for Messages API

# JIRA Ticket
https://nylas.atlassian.net/browse/AV-2664

# License

### DIFF
--- a/nylas/models/messages.py
+++ b/nylas/models/messages.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass, field
-from dataclasses_json import dataclass_json
-from typing import List, Optional, Literal
+from dataclasses_json import dataclass_json, config
+from typing import List, Literal, Optional
 from typing_extensions import TypedDict, NotRequired
 from datetime import datetime
 
@@ -22,8 +22,8 @@ class Participant:
         email: Participant's email
     """
 
-    name: NotRequired[str]
     email: str
+    name: Optional[str] = None
 
 
 @dataclass_json
@@ -41,8 +41,8 @@ class Attachment:
 
     id: str
     size: int
-    filename: NotRequired[str]
-    content_type: NotRequired[str]
+    filename: Optional[str] = None
+    content_type: Optional[str] = None
 
 
 @dataclass_json
@@ -76,14 +76,9 @@ class Message:
     thread_id: str
     subject: str
 
-    from_: List[Participant] = field(
-        metadata={"dataclasses_json": {"field_name": "from"}}
-    )
+    from_: List[Participant] = field(metadata=config(field_name="from"))
     to: List[Participant]
-    cc: List[Participant]
-    bcc: List[Participant]
 
-    reply_to: Optional[List[Participant]]
     date: datetime
 
     unread: bool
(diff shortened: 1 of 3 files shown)


spang:
Unblocking my pet v3 side projects again! This could use a bit more testing—I've only verified that listing messages with `any_email` and `from` filters works so far. I'll continue to run it through its paces but I figured it's worth putting up a PR early so you know it exists. Figuring out how to properly deal with attributes that have the same names as Python keywords ("from", "in") was "fun"!

nylas/models/messages.py
```diff
@@ -0,0 +1,179 @@
...
+        "starred": NotRequired[bool],
+        "thread_id": NotRequired[str],
+        "received_before": NotRequired[int],
+        "received_after": NotRequired[int],
+        "has_attachment": NotRequired[bool],
+        "fields": NotRequired[Fields],
+        "search_query_native": NotRequired[str],
+    },
+)
+"""
```
spang, line 121: FYI, I have no idea if this docstring will end up in the autogenerated docs. Couldn't figure out any way to include a docstring on a functionally declared TypedDict.

nylas/models/messages.py
```diff
@@ -0,0 +1,179 @@
...
+    folders: NotRequired[List[str]]
+
+
+class UpdateMessageRequest:
+
+    """Request payload for updating a message."""
+
+    unread: NotRequired[bool]
+    starred: NotRequired[bool]
+    folder: NotRequired[List[str]]
```
mrashed-dev, line 179: I don't think there's any query/filter params for updating a message. For the `UpdateMessagRequest` the class hould also have `metadata` as a field that can be update and be a TypedDict:

```suggestion
class UpdateMessageRequest(TypedDict):

    """
    Request payload for updating a message.
    
    Attributes:
        starred: The message's starred status
        unread: The message's unread status
        folders: The message's folders
        metadata: A list of key-value pairs storing additional data
    """

    unread: NotRequired[bool]
    starred: NotRequired[bool]
    folder: NotRequired[List[str]]
    metadata: NotRequired[Dict[str, Any]]
```

nylas/resources/messages.py
```diff
@@ -0,0 +1,106 @@
...
+            response_type=Message,
+            query_params=query_params,
+        )
+
+    def update(
+        self,
+        identifier: str,
+        message_id: str,
+        request_body: UpdateMessageRequest,
+        query_params: UpdateMessageQueryParams,
```
mrashed-dev, line 68: Same as above, I don't think there's any query/filter params for updating a message

mrashed-dev (changes requested):
Thanks for this PR @spang 💪🏼 Left some PR comments!

spang:
Alright @mrashed-dev, I think this is ready to merge! I've tested all the basic calls and a subset of query params manually from a script—IMO full/comprehensive testing should be done by a dedicated test rig for the SDK or SDKs and operate off of the API spec.

Also, just pointing out that this PR doesn't include send support—so no C, just R-U-D! Send can follow in a separate PR.

mrashed-dev (approved):
@spang thank you so much for the PR & changes! Message Send + Drafts + Threads + Schedule Send + Smart Compose is currently being worked on, so soon we will have full messages support on the Python SDK 😄

PR #309, opened 2023-11-03
Make email to optional

atejada:
# License

When sending an email without a To (using bc or bcc), fetching breaks.

https://nylas.atlassian.net/jira/software/c/projects/AV/boards/125?selectedIssue=AV-2705

### DIFF
--- a/nylas/models/messages.py
+++ b/nylas/models/messages.py
@@ -78,7 +78,6 @@ class Message:
     subject: str
 
     from_: List[EmailName] = field(metadata=config(field_name="from"))
-    to: List[EmailName]
 
     date: datetime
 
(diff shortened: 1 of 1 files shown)


johnjjung:
In what situation would you have to send a message with a to field?

mrashed-dev:
@atejada You are not able to send an email without a To field with the API. If you attempt to send one, you get:

```json
{
    "request_id": "uuid",
    "error": {
        "type": "invalid_request_error",
        "message": "field 'to' should not be empty"
    }
}
```

atejada:
@johnjjung @mrashed-dev Sorry, what I meant is that if you open your Gmail and send an email without a **To** (you use bc or bcc...I test this already), when you read it using the Python SDK it will break as the **To** is empty.

mrashed-dev (approved):
Thanks for finding + fixing @atejada !! 🚀

PR #310, opened 2023-11-04
Add Message Send, Drafts, Threads, and Smart Compose APIs support

mrashed-dev:
# Description
This PR builds on #305. Adds the remaining message-related endpoints and models. This PR also brings forward a few breaking changes that alter behaviour to #305:
- In `ListMessageQueryParams` anything that was a list type in the API (to, bcc, etc.) are now List type as well. We then convert these arrays to comma delimited strings fit for the API.
- `Attachments` have moved to `models/attachments.py` as we needed a common space for both `Message` and `Draft` objects to pull from.

# License


nylas/models/attachments.py
```diff
@@ -0,0 +1,47 @@
...
+        content_disposition: The content disposition of the attachment.
+        is_inline: Whether the attachment is inline.
+    """
+
+    filename: str
+    content_type: str
+    content: str
+    size: int
+    content_id: NotRequired[str]
+    content_disposition: NotRequired[str]
```
spang, line 46: I'm curious—is `content_disposition` part of the schema for attachment objects, or just part of these creation requests? Have you tested what fields the API sends back when fetching a message with an attachment? (I didn't)

nylas/models/attachments.py
```diff
@@ -0,0 +1,47 @@
...
+    is_inline: NotRequired[bool]
```
spang, line 47: hm, same for this one!

nylas/models/drafts.py
```diff
@@ -0,0 +1,162 @@
...
+        unread (Optional[bool]): Whether the message is unread.
+        starred (Optional[bool]): Whether the message is starred.
+        snippet (Optional[str]): A snippet of the message body.
+        body (Optional[str]): The body of the message.
+        attachments (Optional[List[Attachment]]): The attachments on the message.
+        folders (Optional[List[str]]): The folders that the message is in.
+        created_at (Optional[int]): Unix timestamp of when the message was created.
+    """
+
+    object: str = "draft"
```
spang, line 41: should this on other object types too, like attachments? I see you added it to messages

nylas/models/messages.py
```diff
@@ -70,30 +53,28 @@ class Message:
...
-
     date: datetime
-
-    unread: bool
-    starred: bool
-
-    snippet: str
-    body: str
-
+    object: str = "messages"
```
spang, line 63: should this be `message` or is it actually plural? :thinking:

nylas/models/drafts.py
```diff
@@ -0,0 +1,162 @@
...
+    starred: NotRequired[bool]
+    send_at: NotRequired[int]
+    reply_to_message_id: NotRequired[str]
+    tracking_options: NotRequired[TrackingOptions]
+
+
+UpdateDraftRequest = CreateDraftRequest
+""" A request to update a draft. """
+
+
```
spang, line 101: might be worth commenting here about the functional declaration being because of python keywords "from" and "in" for future readers if they don't read the messages models first

nylas/models/drafts.py
```diff
@@ -0,0 +1,162 @@
...
+        This value should be taken from a ListResponse object's next_cursor parameter.
+"""
+
+
+class SendMessageRequest(CreateDraftRequest):
+    """
+    A request to send a message.
+
+    Attributes:
+        thread_id (NotRequired[str]): The thread that this message belongs to.
```
spang, line 145: do we want to consistently declare types in the docstrings? just noticing this is very inconsistent throughout the SDK

nylas/models/messages.py
```diff
@@ -103,12 +84,12 @@ class Message:
     {
         **get_type_hints(ListQueryParams),  # Inherit fields from ListQueryParams
         "subject": NotRequired[str],
-        "any_email": NotRequired[str],
-        "from": NotRequired[str],
-        "to": NotRequired[str],
-        "cc": NotRequired[str],
-        "bcc": NotRequired[str],
-        "in": NotRequired[str],
+        "any_email": NotRequired[List[str]],
```
spang, line 87: thanks for fixing this stuff!

nylas/resources/messages.py
```diff
@@ -85,24 +123,99 @@ def update(
...
+            Response: The list of scheduled messages.
+        """
+        json_response = self._http_client._execute(
+            method="GET",
+            path=f"/v3/grants/{identifier}/messages/schedules",
+        )
+
+        return Response.from_dict(json_response, ScheduledMessagesList)
+
+    def find_scheduled_messages(
```
spang, line 183: should this be `find_scheduled_message`? given it only returns one

nylas/resources/messages.py
```diff
@@ -85,24 +123,99 @@ def update(
...
+            Response: The scheduled message.
+        """
+        json_response = self._http_client._execute(
+            method="GET",
+            path=f"/v3/grants/{identifier}/messages/schedules/{schedule_id}",
+        )
+
+        return Response.from_dict(json_response, ScheduledMessage)
+
+    def stop_scheduled_messages(
```
spang, line 203: same here—`stop_scheduled_message`?

spang (approved):
I left a few smaller comments inline, but generally looks good to me! Excited to see these features land.

I didn't do a detailed cross reference of parameters and schemas with the docs. Do you have a better, automated way to do that vs manually cross-checking?

nylas/models/attachments.py
mrashed-dev: Yup, it exists on creation and retrieval. Gere's an example response from the `GET /messages` endpoint with an attachment:

```json
{
    "id": "snip",
    "grant_id": "snip",
    "filename": "invite.ics",
    "size": 2420,
    "content_type": "application/ics; name=\"invite.ics\"",
    "is_inline": false,
    "content_disposition": "attachment; filename=\"invite.ics\""
}
```

Note that `is_inline` is also present.

nylas/models/attachments.py
mrashed-dev: See answer above please!

nylas/models/drafts.py
mrashed-dev: `attachments` doesn't need one because this `object` key typically only exists on top-level objects with getter endpoints. Like Drafts, Calendars, Messages, Events, etc. Just tested for attachment objects within messages and drafts, no object key is present.

nylas/models/drafts.py
mrashed-dev: So here's the thing - docstrings kinda suck for more complex applications, such as inheritance. If we don't mention the fields (and their types) of the superclass, it does not automatically inherit it in the docs, so in cases like this, we need to manually add it (and its types).

On top-level objects, all we need to do is mention the field name, and the docstring will be smart enough to take the type of the field from the code directly. This issue is strictly with sub-classes.

nylas/models/messages.py
mrashed-dev: Great catch, will fix. Thanks!

nylas/resources/messages.py
mrashed-dev: You would be correct! Will fix it, thank you :)

nylas/resources/messages.py
mrashed-dev: You would be correct, again! Will also fix it, thank you :)

nylas/models/drafts.py
mrashed-dev: Good idea, adding it!

mrashed-dev:
Changes added, thanks for your thorough review @spang! 

Regarding:
> Do you have a better, automated way to do that vs manually cross-checking?

At the moment, no. We're hoping that, post-GA, we can leverage the OpenAPI spec for the API to see if we can do any sort of codegen whether it's just for documentation but for models as well. The problem is the OpenAPI spec is not 100% accurate with how much development is going on the API side.

PR #319, opened 2023-12-11
Fix issue with multipart attachments throwing KeyError

mrashed-dev:
# Description
We were attaching files via MultipartEncoder the wrong way. It's not expecting raw data, we needed to pass in the fields as a tuple instead of a dict.

# License


atejada (approved):
Just test it and works perfectly 🥳

PR #327, opened 2024-01-03
Fix Invalid Escape Sequence for Image Extraction Regex

max-muoto:
# License

When running Python 3.12, the invalid escape sequence here leads to a `SyntaxWarning`.


mrashed-dev (approved):
Thanks for opening the PR @max-muoto!

PR #335, opened 2024-01-25
Update events.py to add capacity

atejada:
Add capacity to the events response

https://nylas.atlassian.net/browse/TW-2502

# License

### DIFF
--- a/nylas/models/events.py
+++ b/nylas/models/events.py
@@ -282,7 +282,6 @@ class Event:
         id: Globally unique object identifier.
         grant_id: Grant ID representing the user's account.
         calendar_id: The Event's Calendar ID.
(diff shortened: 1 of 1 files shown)


nylas/models/events.py
```diff
@@ -309,6 +310,7 @@ class Event:
     id: str
     grant_id: str
     calendar_id: str
+    capacity: int
```
mrashed-dev, line 313: Capacity should be optional, can you mark it as so and move it down with the optional fields? Thanks!

mrashed-dev (changes requested):
Thanks @atejada ! Left a comment.

atejada:
@mrashed-dev Should be done 😁

mrashed-dev (approved):
Thanks @atejada !

PR #364, opened 2024-05-15
Added support for event_type filtering field for listing events

mrashed-dev:
# Description
This PR adds support for filtering events by event type. Note that this feature is only for Google events.

# License


atejada (approved):
Just test it 👍 Let's 🚀 and BTW...

```
Focus time is only available for eligible work or school accounts.
```

I was breaking my head looking for it on my personal account 😅

PR #369, opened 2024-05-17
Fix deserialization error with event participant

mrashed-dev:
# Description
Event participant status is optional, as the field doesn't get populated for iCloud events.

# License


kraju3 (approved):
LGTM

PR #382, opened 2024-09-05
Make from field Optional to not cause Key errors

kraju3:
# Description
https://nylas.zendesk.com/agent/tickets/94722

# License

### DIFF
--- a/nylas/models/messages.py
+++ b/nylas/models/messages.py
@@ -56,7 +56,7 @@ class Message:
     """
 
     grant_id: str
-    from_: List[EmailName] = field(metadata=config(field_name="from"))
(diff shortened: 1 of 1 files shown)


AaronDDM:
@kraju3 is this expected API behaviour...? seems odd, doesn't it? Or does it happen for drafts?

kraju3:
@AaronDDM It seems like Microsoft Graph API sometimes omit this value from their response and we won't translate it on our end. Since the field is not typed to optional our SDK throws a KeyError while decoding and users won't be able to process the response via the SDK. Making it Optional will solve the KeyError so that they can at least enrich the data.

AaronDDM (approved):

PR #383, opened 2024-09-09
Fix IMAP identifiers not encoding correctly

mrashed-dev:
# Description
IMAP identifiers sometimes have characters that do not work well in the URL, including slashes, that would make the request fail. For these endpoints we ensure that we are encoding these IDs properly.

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 nylas-python Changelog
 ======================
 
+Unreleased
+----------------
+* Fix IMAP identifiers not encoding correctly
+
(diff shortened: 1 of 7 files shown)


kraju3 (approved):
LGTM! I have only noticed these with Email API's do you think we should treat the other resources as well?

mrashed-dev:
@kraju3 I think for now IMAP identifiers are the only problem identifiers, making a clean solution that works across all endpoints will require some rework to how we set the path so I'd want to avoid doing that for now.

PR #385, opened 2024-09-10
Add missing schedule-specific fields to Message model

mrashed-dev:
# Description
This PR adds in two fields, `schedule_id` and `send_at`, to the `Message` model. These fields are returned by the API if the message being sent is scheduled.

# License


kraju3 (approved):
LGTM

PR #396, opened 2024-12-09
Validate attachment response and throw error

kraju3:
# Description
At the moment we are not returning a NylasAPIError exception if we are unable to download the attachment. Which results in customers writing the content to a file. This PR adds a validation for the download request and return an exception if the status code is >=400. 

# License

### DIFF
--- a/nylas/handler/http_client.py
+++ b/nylas/handler/http_client.py
@@ -45,11 +45,6 @@ def _validate_response(response: Response) -> dict:
 
     return json
 
-def _validate_download_response(response:Response) -> Union[bytes, Response, dict]:
-    if response.ok:
-        return response.content
-    return _validate_response(response)
-
 def _build_query_params(base_url: str, query_params: dict = None) -> str:
     query_param_parts = []
     for key, value in query_params.items():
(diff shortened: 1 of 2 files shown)


nylas/handler/http_client.py
```diff
@@ -126,9 +130,9 @@ def _execute_download_request(
 
             # If we stream an iterator for streaming the content, otherwise return the entire byte array
             if stream:
-                return response
+                return _validate_download_response(response,stream)
 
-            return response.content if response.content else None
+            return _validate_download_response(response) if response.content else None
```
AaronDDM, line 135: @kraju3 do we need to make this change here? Looks like if it's `stream` we handle the response differently earlier in this function anyways.

nylas/handler/http_client.py
```diff
@@ -126,9 +130,9 @@ def _execute_download_request(
 
             # If we stream an iterator for streaming the content, otherwise return the entire byte array
             if stream:
-                return response
+                return _validate_download_response(response,stream)
```
AaronDDM, line 133: Maybe we can simplify this?

```suggestion
                return response if response.ok else response.content
```

AaronDDM: But then, will `_execute_download_request` not error out where ever it is used 🤔 So I guess the question is, is this the correct place to do this validate this?

nylas/handler/http_client.py
kraju3: Hey @AaronDDM I was hoping to keep the same behavior plus return an error if the error code is different

kraju3:
@AaronDDM Is it good to merge this PR?

AaronDDM (approved):

PR #404, opened 2025-01-27
Added response headers to all responses from the Nylas API

AaronDDM:
# Background
Nylas has added a new `retry-header` that will be included as part of our error responses. This PR exposes these response headers for all success and error responses.

# License


AaronDDM:
> I only see some resource endpoints. Do we have support for this in threads endpoints?

Because threads uses the underlying resources class, it automatically gets this support. Files that didn't do that needed to be updated individually.

kraju3 (approved):

PR #406, opened 2025-01-31
Add webhook triggers to bypass ValueErrors

kraju3:
# Description
We were missing some webhook triggers which caused ValueErrors while customers tried to use our Webhooks API via the SDK.

# License


AaronDDM (approved):
Are there any tests that need to be updated?

kraju3:
@AaronDDM There is deserialization test but I just added one that covers the entire list

PR #424, opened 2025-06-17
CUST-4514 added handling for non-json responses and tests

samuelpx:
# License


nylas/models/errors.py
```diff
@@ -70,7 +70,7 @@ class NylasApiErrorResponse:
         error: The error data.
     """
 
-    request_id: str
+    request_id: Optional[str]
```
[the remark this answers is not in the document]
samuelpx: Thanks, I'll correct this!

nylas/models/errors.py
```diff
@@ -169,3 +169,47 @@ def __init__(self, url: str, timeout: int, headers: Optional[CaseInsensitiveDict
         self.url: str = url
         self.timeout: int = timeout
         self.headers: CaseInsensitiveDict = headers
+
+
+class NylasNetworkError(AbstractNylasSdkError):
```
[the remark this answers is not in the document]
samuelpx: Nowhere yet, the idea is that it will replace the proposed non-json handling implementation in the future, we currently raise a NylasApiError to keep backwards compatibility as a priority, let me know if we should just use this instead.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ v6.11.0
 
 v6.10.0
 ----------------
+* Added handling for non-JSON responses
```
AaronDDM, line 25: This entry is under `v6.10.0`, which has already been released. Please move it under an `Unreleased` section at the top so it lands in the next release notes.

nylas/handler/http_client.py
```diff
@@ -18,7 +18,31 @@
...
+            flow_id = response.headers.get("x-fastly-id", "")
+            flow_info = f" (flow_id: {flow_id})" if flow_id else ""
+            raise NylasApiError(
+                NylasApiErrorResponse(
+                    "",
+                    NylasApiErrorResponseData(
+                        type="network_error",
+                        message=f"""
+                        HTTP {response.status_code}: Non-JSON response received{flow_info}.
+                        Body: {body_preview}""",
```
AaronDDM, line 39: The triple-quoted f-string inlines the source indentation into the error message, so the rendered text looks like:

```

                        HTTP 500: Non-JSON response received (flow_id: ...).
                        Body: ...
```

That'll look awful in logs and stack traces. Either collapse to a single line:

```python
message=f"HTTP {response.status_code}: Non-JSON response received{flow_info}. Body: {body_preview}"
```

or use `textwrap.dedent` + `.strip()`. The existing tests assert on the indented string, so they'll need updating too — ideally to substring assertions.

nylas/handler/http_client.py
```diff
@@ -18,7 +18,31 @@
...
+                    ),
+                ),
+                status_code=response.status_code,
+                headers=response.headers,
+            ) from exc
+        return ({}, response.headers)
```
AaronDDM, line 45: Silently returning `({}, headers)` for a 2xx non-JSON body could mask a real problem (e.g. a malformed JSON body that happens to come back with 200). At minimum add a debug log; better, consider raising for unexpected non-JSON success too, since every JSON endpoint in this SDK should return parseable JSON.

nylas/models/errors.py
```diff
@@ -28,9 +28,9 @@ def __init__(
             status_code: The HTTP status code of the error response.
             message: The error message.
         """
-        self.request_id: str = request_id
-        self.status_code: int = status_code
-        self.headers: CaseInsensitiveDict = headers
+        self.request_id: Optional[str] = request_id
+        self.status_code: Optional[int] = status_code
+        self.headers: Optional[CaseInsensitiveDict] = headers
```
AaronDDM, line 33: Per the earlier review thread — these shouldn't become Optional on the base class. Genuine API errors always have a `status_code` and `headers`. Keep them required here and let `NylasNetworkError` declare its own Optionals if needed.

nylas/models/errors.py
```diff
@@ -169,3 +169,47 @@ def __init__(self, url: str, timeout: int, headers: Optional[CaseInsensitiveDict
...
+            raw_body: The non-JSON response body.
+            headers: The headers returned from the server.
+            flow_id: The value from x-fastly-id header if present.
+        """
+        super().__init__(message)
+        self.request_id: Optional[str] = request_id
+        self.status_code: Optional[int] = status_code
+        self.raw_body: Optional[str] = raw_body
+        self.headers: Optional[CaseInsensitiveDict] = headers
+        self.flow_id: Optional[str] = flow_id
```
AaronDDM, line 215: This class is defined but never instantiated. Either raise it from `_validate_response` instead of (or alongside) `NylasApiError`, or drop it from this PR and add it when the v7.0 work lands. Right now it's just dead code and a maintenance trap.

tests/handler/test_http_client.py
```diff
@@ -432,3 +432,127 @@ def test_execute_with_headers(self, http_client, patched_version_and_sys, patche
...
+
+    def test_validate_response_500_error_html(self):
+        response = Mock()
+        response.status_code = 500
+        response.json.side_effect = ValueError("No JSON object could be decoded")
+        response.text = "<html><body><h1>Internal Server Error</h1></body></html>"
+        response.headers = {"Content-Type": "text/html", "x-fastly-id": "fastly-123"}
+
+        with pytest.raises(NylasApiError) as e:
+            _validate_response(response)
```
AaronDDM, line 444: These `assert str(e.value) == """..."""` checks pin the exact whitespace of the error string and will all break once the multi-line f-string is fixed. Recommend asserting on substrings, e.g.:

```python
assert "HTTP 500" in str(e.value)
assert "flow_id: fastly-123" in str(e.value)
assert "<h1>Internal Server Error</h1>" in str(e.value)
```

AaronDDM (changes requested):
Solid direction — handling non-JSON 4xx/5xx responses fixes a real pain point. A few things to address before merging:

1. **CHANGELOG entry is under `v6.10.0` (already released).** Move it to an `Unreleased` section so it shows up in the next release notes.
2. **Multi-line f-string in the error message produces ugly literal indentation** (24 spaces of leading whitespace on each subsequent line). See inline comment.
3. **`NylasNetworkError` is declared but never raised anywhere.** Either wire it up or remove it; dead code is confusing (this was flagged back in June and is still unresolved).
4. **`AbstractNylasSdkError` fields made Optional** — already flagged in the earlier review. `request_id`/`status_code`/`headers` are not optional for genuine API errors; weakening these types just to satisfy the network-error path regresses type safety for the main flow. Prefer keeping them required on `AbstractNylasSdkError` and let `NylasNetworkError` (once wired up) override with Optional.
5. **Tests assert exact whole-message strings**, which couples them to the indentation-bug formatting. Once #2 is fixed every test will break; assert on substrings (status code, `flow_id`, body excerpt) instead.
6. **Silent `return ({}, response.headers)` on non-JSON 2xx** — minor, but worth a debug log so this doesn't silently mask a malformed-but-successful response.

PR #428, opened 2025-07-07
Fix grants.update() to use PATCH not PUT

drpancake:
The API only allows PATCH, so an error is always thrown.

# License


samuelpx:
@drpancake Thank you for raising this, James, we really appreciate you taking the time to do so! Moved the fix to https://github.com/nylas/nylas-python/pull/435, tests and other CI had to be updated as well, closing this PR now.

PR #440, opened 2025-09-24
[CUST-4448] Fix: content id not being respected for large inline attachments.

AaronDDM:
# What did you do?

- [x] Updated `_build_form_request` to use `content_id` if exists, otherwise fallback to indexed file# 

# License


samLRodrigues (approved):
LGTM

PR #471, opened 2026-04-24
Add live E2E coverage for policies, rules, and lists

pengfeiye:
# License

# Summary:
- Added live, env-gated E2E test suite under `tests/e2e` for `lists`, `rules`, and `policies`.
- Implemented deterministic cleanup via shared resource registry/finalizers in `tests/e2e/conftest.py` to avoid orphaned resources.
- Covered full lifecycle flows:
  - Lists: create, find, update, add/list/remove items, destroy
  - Rules: create, find, update, list verification, destroy
  - Policies: create, find, update (including rule association), list verification, destroy
- Hardened tests for real API response shapes by:
  - verifying policy updates with follow-up `find` when update payload omits `id`
  - using normalized/raw list ID extraction for list endpoints with variable payload structure
- Registered pytest marker in `pyproject.toml`:
  - `e2e: marks tests that call live Nylas APIs`
- Validation:
  - `pytest tests/e2e -m e2e` passes (3/3)
  - existing unit resource tests remain green

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -1,0 +1,104 @@
+import os
+from typing import Dict, List
+from uuid import uuid4
+
+import pytest
+
+from nylas import Client
+
+
+_API_KEY_ENV_VARS = ("NYLAS_E2E_API_KEY", "NYLAS_API_KEY")
+_API_URI_ENV_VARS = ("NYLAS_E2E_API_URI", "NYLAS_API_URI")
+
+
+def _first_env_value(keys: tuple) -> str:
+    for key in keys:
+        value = os.getenv(key)
+        if value:
+            return value
+    return ""
+
+
+def extract_list_items(response_data):
+    """
+    Normalize list endpoint response payloads across API shapes.
+    """
+    if isinstance(response_data, list):
+        return response_data
+
+    if isinstance(response_data, dict):
+        items = response_data.get("items")
+        if isinstance(items, list):
+            return items
+
+    return []
+
+
+def raw_list_ids(client: Client, path: str, id_key: str = "id", query_params=None):
+    json_response, _headers = client.http_client._execute(
+        "GET",
+        path,
+        None,
+        query_params or {"limit": 200},
+        None,
+    )
+    response_data = json_response.get("data")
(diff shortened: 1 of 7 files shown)


tests/e2e/conftest.py
```diff
@@ -0,0 +1,104 @@
...
+from typing import Dict, List
+from uuid import uuid4
+
+import pytest
+
+from nylas import Client
+
+
+_API_KEY_ENV_VARS = ("NYLAS_E2E_API_KEY", "NYLAS_API_KEY")
+_API_URI_ENV_VARS = ("NYLAS_E2E_API_URI", "NYLAS_API_URI")
```
cgeene: [P2] Live E2E runs can be triggered by normal test commands

The fixture accepts the generic `NYLAS_API_KEY`, and these tests live under `tests/`, which the repo 's default `python setup.py test` command collects. That means any developer or CI environment with a normal SDK API key set will unexpectedly create/update/delete live policies, rules, and lists during the regular unit suite. I would require the explicit `NYLAS_E2E_API_KEY` only, and ideally also default-exclude `e2e` tests unless an explicit opt-in flag/env var is set.

tests/e2e/conftest.py
```diff
@@ -0,0 +1,104 @@
...
+    json_response, _headers = client.http_client._execute(
+        "GET",
+        path,
+        None,
+        query_params or {"limit": 200},
+        None,
+    )
+    response_data = json_response.get("data")
+    items = extract_list_items(response_data)
+    return {item.get(id_key) for item in items if isinstance(item, dict) and item.get(id_key)}
```
cgeene: [P2] Raw helper bypasses the SDK list methods being covered

`raw_list_ids` calls `client.http_client._execute` directly, so the policies/rules list checks can pass even if `e2e_client.policies.list()` or `e2e_client.rules.list()` is broken against the live API. It also only reads one page, which can flake in accounts with more than 200 resources. Since this PR is specifically adding SDK E2E coverage, the tests should use the public resource methods and page through `next_cursor` until the created id is found.

cgeene (approved):

PR #480, opened 2026-07-08
[Python SDK] Add resources field to Event, CreateEventRequest, and UpdateEventRequest models (CUST-5372)

pengfeiye:
## Summary

Adds support for the `resources` field on events (used for room/resource booking) to the Python SDK, resolving CUST-5372 (child of CUST-5262). The Nylas API already supports this field; this change exposes it in the SDK models so resource data is deserialized from responses and can be sent on create/update requests.

## Changes

- **`nylas/models/events.py`**
  - Added a `Resource` response dataclass (`@dataclass_json`) with optional fields: `email`, `name`, `capacity`, `building`, `floor_name`, `floor_section`, `floor_number`.
  - Added `resources: Optional[List[Resource]]` to the `Event` response model so resource data is deserialized from API responses.
  - Added a `WritableResource` `TypedDict` (all keys `NotRequired`) for request bodies, following the existing `Participant`/`CreateParticipant` pattern; reused across both request models since the create/update shape is identical.
  - Added `resources: NotRequired[List[WritableResource]]` to `CreateEventRequest` and `UpdateEventRequest`.
  - Updated docstrings on all affected models.

- **`tests/resources/test_events.py`**
  - Extended `test_event_deserialization`, `test_create_event`, and `test_update_event` to cover the `resources` field.
  - Added edge-case tests: missing `resources`, empty list, partially-populated resource, multiple resources, and create/update request passthrough for empty/partial resources.

## Test plan

- [x] `pytest tests/resources/test_events.py` (28 passed)
- [x] No linter errors on edited files

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ nylas-python Changelog
 Unreleased
 ----------
 * Clarify that event `default` visibility is Google-only
+* Added `resources` field to Event, CreateEventRequest, and UpdateEventRequest models
 
 v6.16.0
 ----------
--- a/nylas/models/events.py
+++ b/nylas/models/events.py
@@ -49,6 +49,31 @@ class Participant:
     phone_number: Optional[str] = None
 
 
+@dataclass_json
+@dataclass
+class Resource:
+    """
+    Class representing an Event resource, such as a bookable room or piece of equipment.
+
+    Attributes:
+        email: The resource's email address.
+        name: The resource's name.
+        capacity: The maximum number of people the resource can accommodate.
+        building: The building the resource is located in.
+        floor_name: The name of the floor the resource is located on.
+        floor_section: The section of the floor the resource is located in.
+        floor_number: The number of the floor the resource is located on.
+    """
+
+    email: Optional[str] = None
+    name: Optional[str] = None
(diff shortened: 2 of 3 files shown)


AaronDDM (changes requested):
Good morning!

The models, docstrings, and test coverage here are great — I especially like the edge cases (missing / empty / partial / multiple resources, plus the create/update passthrough).

One blocker before merge:
- **No CHANGELOG.md entry.** The `Unreleased` section only has the existing "default visibility" line. Please add something like:
  `* Added resources field to Event, CreateEventRequest, and UpdateEventRequest models`

Minor, non-blocking:
- The existing code uses separate `CreateParticipant` / `UpdateParticipant` types, but you've reused a single `WritableResource` for both create and update. Fine while the shapes are identical — just flagging in case they ever diverge.

Requesting changes only for the changelog; everything else looks good. Thanks!

AaronDDM (approved):
Good afternoon! Re-reviewed after the latest push. The `Unreleased` CHANGELOG entry is in now, which was my only blocker last time. Models, docstrings, and the edge-case test coverage (missing / empty / partial / multiple resources, plus create/update passthrough) all look great. Approving — thanks!

PR #481, opened 2026-07-15
Add tentative_as_busy field to GetFreeBusyRequest

pengfeiye:
## Summary

Adds an optional `tentative_as_busy: NotRequired[bool]` field to the `GetFreeBusyRequest` TypedDict in `nylas/models/free_busy.py`. This field already existed on the Events query params and Availability models, but was missing from the free-busy request.

When set to `false`, tentative calendar events are treated as `busy: false`. Only applicable for Microsoft and EWS calendar providers; defaults to `true`.

## Changes

- `nylas/models/free_busy.py`: import `NotRequired` from `typing_extensions` and add `tentative_as_busy: NotRequired[bool]` to `GetFreeBusyRequest`, with a matching docstring entry.
- `tests/resources/test_calendars.py`: add `test_get_free_busy_with_tentative_as_busy` verifying the field is passed through in the request body.

## Testing

- `python -m pytest tests/resources/test_calendars.py -k "free_busy"` — 2 passed.

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Unreleased
 ----------
 * Clarify that event `default` visibility is Google-only
 * Added `resources` field to Event, CreateEventRequest, and UpdateEventRequest models
+* Added tentative_as_busy field to GetFreeBusyRequest
 
 v6.16.0
 ----------
--- a/nylas/models/free_busy.py
+++ b/nylas/models/free_busy.py
@@ -2,7 +2,7 @@ from dataclasses import dataclass
 from typing import List, Union
 
 from dataclasses_json import dataclass_json
-from typing_extensions import TypedDict
+from typing_extensions import TypedDict, NotRequired
 
 
 @dataclass_json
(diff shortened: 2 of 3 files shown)


AaronDDM (changes requested):
Good afternoon! Thanks for this — clean change and the test covers the request passthrough nicely.

One blocker before merge:
- **No CHANGELOG.md entry.** Please add a line under `Unreleased`, e.g. `* Added tentative_as_busy field to GetFreeBusyRequest`.

The field, docstring, and `NotRequired[bool]` typing all look correct and match how `tentative_as_busy` is handled on the other event/availability models. Requesting changes only for the changelog — happy to approve once that's in.

AaronDDM (approved):
Good afternoon! Thanks for the quick turnaround — the CHANGELOG entry under Unreleased is in now, which was the only blocker from my last pass.

Everything else still looks good: the `NotRequired[bool]` typing, the docstring, and the test covering the request passthrough are all consistent with how `tentative_as_busy` is handled on the other models. Approving. 🚀

PR #483, opened 2026-07-27
fix(webhooks): forward overrides from webhooks.list

toolshedlabs-hash:
# License

## The defect

`client.webhooks.list(overrides=...)` accepts the argument and documents it, then calls the
base method without it. The request goes out with the client's own settings, so a per
request `api_key`, `api_uri`, `timeout` or `headers` has no effect on that one call.

`ListableApiResource.list` accepts `overrides` and forwards it to `_execute`, so nothing
further down is putting it back.

## Reproduction

Clean venv, `nylas==6.17.0` from PyPI, Python 3.14.0. One client, one loopback server that
records the `Authorization` header of every request.

```python
client = Client(api_key="CLIENT-DEFAULT-KEY", api_uri=f"http://127.0.0.1:{PORT}")
OVERRIDES = {"api_key": "PER-REQUEST-OVERRIDE-KEY"}
```

| call | request | Authorization the server received |
| --- | --- | --- |
| `webhooks.list(overrides=OVERRIDES)` | GET /v3/webhooks | `Bearer CLIENT-DEFAULT-KEY` |
| `webhooks.find("wh_1", overrides=OVERRIDES)` | GET /v3/webhooks/wh_1 | `Bearer PER-REQUEST-OVERRIDE-KEY` |
| `webhooks.ip_addresses(overrides=OVERRIDES)` | GET /v3/webhooks/ip-addresses | `Bearer PER-REQUEST-OVERRIDE-KEY` |
| `grants.list(overrides=OVERRIDES)` | GET /v3/grants | `Bearer PER-REQUEST-OVERRIDE-KEY` |

The last row is the control worth having. It is a `list` on a different resource, so the
shape forwards overrides fine and the problem is this one method.

## How wide it is

An AST pass over `nylas/resources/` finds 141 methods that accept `overrides`, and checks
whether each one reads the name anywhere in its body. Exactly one never does, and this is
it. The check counts positional and renamed forwarding as correct, which matters, because
`auth.py` passes it positionally as `self._get_token(request_body, overrides)` and
`domains.py` passes it as `overrides=merged` after `_merge_signer_headers` runs.

## The change

```diff
-        return super().list(path="/v3/webhooks", response_type=Webhook)
+        return super().list(
+            path="/v3/webhooks",
+            response_type=Webhook,
+            overrides=overrides,
+        )
```

Plus a regression test in `tests/resources/test_webhooks.py` that asserts `_execute` is
called with the overrides, and a CHANGELOG bullet. The test fails on main and passes with
the change.

`pytest` is 475 passed, 3 deselected. `pylint nylas/resources/webhooks.py` is 10.00.

The CHANGELOG bullet went under a new `Unreleased` heading because there is not one at the
moment. Happy to drop that hunk if you would rather add it at release time.


toolshedlabs-hash:
Closing this. It has picked up a merge conflict and nobody has needed it in two weeks. Happy to rebase and reopen if it is still wanted.

